PR #2, opened 2025-10-29
WEBDEV-7899 Add UX team components to README

rebecca-shoptaw:
Adds a (non-exhaustive) list of UX-team-created components are already or easily-could-be generalized, each with a relevant `Point` person, as applicable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,11 +78,28 @@ have built thus far for their respective teams.
 - __ia-otp-input (UX team)__ - Complex little component to handle rendering a 6-box OTP input. Autofocuses first input, hops between inputs, supports pasting, autocomplete, a custom number of boxes and a custom allowed-characters regex (defaults to numeric-only). Used i.e. in the `Email me a code` modal on the [account settings](https://archive.org/account/settings) page. Point: [@rebecca-shoptaw][rebecca-shoptaw]. [(link)](https://git.archive.org/www/offshoot/-/blob/main/src/widgets/otp/ia-otp-input.ts)
 - __ia-clearable-text-input__ - Text input with a clear button and substantial a11y support. Point: [@latonv][latonv]. [(link)](https://github.com/internetarchive/iaux-clearable-text-input/blob/main/src/ia-clearable-text-input.ts)
 
+### Menu
+
+- __ia-dropdown__ - Supports various customized dropdown menu use cases. Point: [@latonv][latonv]. [(link)](https://github.com/internetarchive/iaux-dropdown)
+
+### Infinite Scroller
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,19 +38,52 @@ have built thus far for their respective teams.
 
 - __ads-card (vault)__ - Very basic card component, in use in Vault a couple places.
   [(link)](https://git.archive.org/dps/vault-site/-/blob/fe99de1d53c62f426b1d6b70cc100321d651a9aa/frontend/src/web-components/ads-card.ts)
+- __ia-image-card__ - Simple component to display an image as a card, with an optional overlay color, hover-title, and on-click link. Used by various custom collections such as [Democracy's Library](https://archive.org/details/democracys-library). Point: @rebecca-shoptaw. [(link)](https://git.archive.org/www/offshoot/-/blob/main/src/widgets/ia-image-card.ts)
 
 ### Toggle Switch
 
 - __ads-toggle-switch (vault)__ - Nice toggle switch with CSS animation moving toggle back and forth.
   [(link)](https://git.archive.org/dps/vault-site/-/blob/fe99de1d53c62f426b1d6b70cc100321d651a9aa/frontend/src/web-components/ads-toggle-switch.ts)
+- __toggle-switch (UX team)__ - Toggle switch with screen-reader support and a lot of CSS customization options. Point: @nsharma123. [(link)](https://git.archive.org/www/offshoot/-/blob/main/src/widgets/toggle-switch.ts)
```
latonv: I could be a point person for this one too -- I think the one in Offshoot was largely copied from the original in https://github.com/internetarchive/iaux-collection-browser/blob/main/src/collection-facets/toggle-switch.ts

README.md
latonv: A few other things that might be worth including if they seem generalizable enough:

- The `ia-dropdown` component [(link)](https://github.com/internetarchive/iaux-dropdown) that supports various customized dropdown menu use cases.
- Our `infinite-scroller` component [(link)](https://github.com/internetarchive/iaux-infinite-scroller) for virtualized buffered scrolling of large lists.
- The histogram component currently living [here](https://git.archive.org/www/offshoot/-/blob/main/src/scenes/collection-page/components/histograms/collection-histogram.ts). Only used on collection pages at the moment, but the component is fairly general already.
- Similarly, the `histogram-date-range` component [(link)](https://github.com/internetarchive/iaux-histogram-date-range), which we might explore consolidating with the above?
- Our `SharedResizeObserver`, which could be made into a more declarative component similar to [what Shoelace has done](https://shoelace.style/components/resize-observer). I.e., the component itself manages attaching/detaching the shared observer instance from its slotted children as part of its lifecycle, so that we don't have to repeat that logic all over our other components.

jeffwklein:
LGTM

README.md
rebecca-shoptaw: Thanks so much!! I'll add these to the list tomorrow morning and have you as a point for all of them, especially the shared resize observer rework which would be great to have.

PR #36, opened 2026-02-27
WEBDEV-8131 Encapsulate property settings

rebecca-shoptaw:
Encapsulates property settings for easy reuse, which can be extended gradually to be used for i.e. `ia-combo-box` and other more complicated components.

Also extracts property settings and style settings out of the `story-template` and encapsulates `exampleUsage` inside template (with the option to override)

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,54 +148,61 @@ Import your story in `/demo/app-root.ts` and add it to the page.
 
 The story template is a component you can use in your story to demo your component. This lets us present them in a consistent way.
 
-It has 5 main configurations:
+It has a few main configurations:
 
 *Properties*
 - `elementTag` (_string_) your component's name, ie `ia-button`
-- `exampleUsage` (_string_) your element's example usage, displays it with syntax highlighting
 - `labs` (_boolean_) if your component is in `labs` to update links
+- `styleInputSettings` (_StyleInputSettings array_) the style options to display, in the appropriate format
+- `propInputSettings` (_PropInputSettings array_) the prop options to display, in the appropriate format
+
+**Note:** If the type of prop input you want to use isn't yet available in the `propInputSettings` options,
+you can instead slot in custom property settings by using `slot="settings"`, adding a button to apply the
+properties, and passing in a `customExampleUsage` (see `ia-snow-story.ts` for a sample implementation).
 
 *Slots*
 - `demo` put your component demo in here
-- `settings` put your component settings in here
 
 Here's an example:
 ```typescript
 import '@demo/story-template';
+
+// Style options to display
+const styleInputSettings: StyleInputSettings[] = [
+  {
+    label: 'Button BG color',
+    cssVariable: '--button-background-color',
+    defaultValue: 'blue',  // Should match the actual fallback value for the variable
+    inputType: 'color',
+  },
+];
+
+// Prop options to display
+const propInputSettings: PropInputSettings[] = [
+  {
+    label: 'Mode',
+    inputType: 'radio',
+    propertyName: 'mode', // Should match the name of the property from its @property() declaration
+    radioOptions: ['primary', 'secondary', 'danger'],
+    defaultValue: 'primary', // Should match the actual default value from the @property() declaration, if any
+  },
+]
 ...
 render() {
   return html`
-    <story-template elementTag="ia-button" elementClassName="IAButton" .exampleUsage=${this.exampleUsage}>
+    <story-template 
+      elementTag="ia-button" 
+      elementClassName="IAButton" 
+      .styleInputSettings=${this.styleInputSettings} 
+      .propInputSettings=${this.propInputSettings}
+    >
       <div slot="demo">
-        <ia-button @click=${() => alert('Button clicked!')}
-          >Click Me</ia-button
-        >
-      </div>
-
-      <div slot="settings">
-        <table>
-          <tr>
-            <td>Color</td>
-            <td><input type="color" value="#007bff" id="color" /></td>
-          </tr>
-        </table>
-        <button @click=${this.apply}>Apply</button>
+        <ia-button @click=${() => alert('Button clicked!')}>
+          Click Me
+        </ia-button>
       </div>
     </story-template>
   `;
-}
-
-// return a string of your element's usage and it will
-// be displayed with syntax highlighting
-// can be dynamic
-private get exampleUsage(): string {
-  return `<ia-button @click=\${() => alert('Button clicked!')}>
-    Click Me
-  </ia-button>`;
-}
-
-private apply(): void {
-  // update component settings based on settings change
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -193,8 +193,8 @@ render() {
     <story-template 
       elementTag="ia-button" 
       elementClassName="IAButton" 
-      .styleInputSettings=${this.styleInputSettings} 
-      .propInputSettings=${this.propInputSettings}
+      .styleInputData=${{settings: this.styleInputSettings}} 
+      .propInputData=${{settings: this.propInputSettings}}
     >
       <div slot="demo">
         <ia-button @click=${() => alert('Button clicked!')}>
--- a/demo/story-components/story-prop-settings.ts
+++ b/demo/story-components/story-prop-settings.ts
@@ -1,0 +1,153 @@
+import {
+  css,
+  html,
+  LitElement,
+  nothing,
+  TemplateResult,
+  type CSSResultGroup,
+} from 'lit';
+import { property, queryAll } from 'lit/decorators.js';
+import { customElement } from 'lit/decorators/custom-element.js';
+import { choose } from 'lit/directives/choose.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
+
+import themeStyles from '@src/themes/theme-styles';
+import { labelToId } from '../story-utils';
+
+export type PropInputSettings<T> = {
+  label: string;
+  propertyName: keyof T;
+  defaultValue?: string;
+  inputType?: 'text' | 'radio';
+  radioOptions?: string[];
+};
+
+export type PropInputData = {
+  settings: PropInputSettings<any>[];
+};
+
+export type AppliedProps = {
+  propName: string;
+  value: string | boolean | number;
+}[];
+
+/**
+ * A template for displaying the prop options.
+ */
+@customElement('story-props-settings')
+export class StoryTemplate extends LitElement {
+  @property({ type: Object }) propInputData?: PropInputData;
+
+  @queryAll('.prop-input')
+  private propInputs?: NodeListOf<HTMLInputElement>;
+
+  render() {
+    if (!this.propInputData) return nothing;
+
+    return html`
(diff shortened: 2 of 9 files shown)


bfalling (commented):
** Redacted Claude-generated single-post code review here in favor of having it post inline **

demo/story-components/story-prop-settings.ts
```diff
@@ -0,0 +1,153 @@
...
+  settings: PropInputSettings<any>[];
+};
+
+export type AppliedProps = {
+  propName: string;
+  value: string | boolean | number;
+}[];
+
+/**
+ * A template for displaying the prop options.
```
bfalling, line 35: **Class name collision:** Both this file and `story-styles-settings.ts` export a class named `StoryTemplate`. While they register different custom element names, this is confusing and will cause issues if anyone tries to import both classes by name. Consider renaming to `StoryPropSettings` and `StoryStyleSettings` respectively.

demo/story-components/story-styles-settings.ts
```diff
@@ -0,0 +1,87 @@
...
+  defaultValue?: string;
+  inputType?: 'color' | 'text';
+};
+
+export type StyleInputData = {
+  settings: StyleInputSettings[];
+};
+
+/**
+ * A template for display the style options.
```
bfalling: Same `StoryTemplate` class name issue as `story-prop-settings.ts` (see comment there). Also, minor typo in JSDoc: "for display the style options" → "for displaying the style options".

demo/story-components/story-prop-settings.ts
```diff
@@ -0,0 +1,153 @@
...
+  propertyName: keyof T;
+  defaultValue?: string;
+  inputType?: 'text' | 'radio';
+  radioOptions?: string[];
+};
+
+export type PropInputData = {
+  settings: PropInputSettings<any>[];
+};
+
```
bfalling, line 28: **Unnecessary wrapping type:** Both `PropInputData` and `StyleInputData` wrap the settings array in `{ settings: T[] }`. Callers then have to write `.propInputData=${{ settings: propInputSettings }}`. Accepting the array directly would be simpler and more consistent with the README examples, which show the arrays being passed directly.

demo/story-template.ts
```diff
@@ -92,7 +93,9 @@ export class StoryTemplate extends LitElement {
         <h3>Usage</h3>
         <syntax-highlighter
           language="auto"
-          .code=${this.exampleUsage}
+          .code=${this.customExampleUsage
```
bfalling: **Simplify ternary:** Since `customExampleUsage` is `string | undefined`, this can be simplified to:
```ts
.code=${this.customExampleUsage ?? this.exampleUsage}
```

demo/story-template.ts
```diff
@@ -163,12 +140,16 @@ import '${this.modulePath}';
     }
   }
 
+  private get exampleUsage(): string {
```
bfalling, line 141: **Dense one-liner:** This generated `exampleUsage` getter is hard to read and maintain. Consider breaking it into a multi-line method for clarity.

demo/story-template.ts
```diff
@@ -163,12 +140,16 @@ import '${this.modulePath}';
...
+    return `<${this.elementTag}${this.defaultUsageProps ? '\n ' + this.defaultUsageProps + '\n' : ''}${this.stringifiedProps ?? ''}></${this.elementTag}>`;
+  }
+
   private get cssCode(): string {
-    if (!this.appliedStyles) return '';
+    if (!this.stringifiedStyles) return '';
     return `
```
bfalling, line 151: **Inconsistent indentation:** This has a single-space indent before `${this.stringifiedStyles}`, whereas the old code used two spaces. This changes the CSS output formatting — likely unintentional.

demo/story-template.ts
```diff
@@ -185,21 +166,32 @@ ${this.elementTag} {
...
-  /* Applies styles to demo component. */
-  private applyStyles(): void {
-    const appliedStyles: string[] = [];
+  /* Detects and stores a reference to slotted demo component */
+  private handleDemoComponentSlotted(e: Event): void {
+    const slottedComponent = (
+      e.target as HTMLSlotElement
+    ).assignedElements()[0];
+    if (slottedComponent) this.slottedDemoComponent = slottedComponent;
+  }
```
bfalling, line 178: **Avoid `any` type:** `slottedDemoComponent` is typed as `any`. Consider typing it as `Element` or `HTMLElement` since that's what `assignedElements()` returns.

demo/story-template.ts
```diff
@@ -185,21 +166,32 @@ ${this.elementTag} {
...
 
-    this.appliedStyles = appliedStyles.join('\n  ');
+    this.stringifiedStyles = stringifiedStyles;
   }
 
-  /* Converts a label to a usable input id, i.e. My setting -> my-setting */
-  private labelToId(label: string): string {
-    return label.toLowerCase().split(' ').join('-');
+  /* Applies props from the settings to the component and code demo */
+  private handlePropsApplied(e: CustomEvent): void {
```
bfalling, line 188: **`setAttribute` vs property assignment:** `setAttribute` sets HTML attributes (always strings), but Lit `@property()` bindings typically use JS properties which can be typed. For non-string properties (booleans, numbers, objects), `setAttribute` won't work correctly. Should use property assignment instead:
```ts
(this.slottedDemoComponent as any)[prop.propName] = prop.value;
```

demo/story-components/story-prop-settings.ts
```diff
@@ -0,0 +1,153 @@
...
+    const inputId = labelToId(settings.label);
+
+    return html`
+      <tr>
+        <td><legend>${settings.label}</legend></td>
+        <td>
+          ${settings.radioOptions.map(
+            (option) =>
+              html`<input
```
bfalling, line 100: **Potential `id` collision on radio inputs:** The `id` attribute is set to `option` (the option value string). If two different prop settings have overlapping radio option values (e.g., both have a `"true"` option), the `id` and `for` attributes will collide, breaking label association. Should prefix with the input group name, e.g. `${inputId}-${option}`.

README.md
```diff
@@ -148,55 +148,62 @@ Import your story in `/demo/app-root.ts` and add it to the page.
...
 
 *Slots*
 - `demo` put your component demo in here
-- `settings` put your component settings in here
 
 Here's an example:
 ```typescript
 import '@demo/story-template';
+
+// Style options to display
```
bfalling, line 170: **README examples don't match actual API.** The README shows `propInputSettings` and `styleInputSettings` being passed directly as properties on `<story-template>`, but the actual component expects them wrapped in `{ settings: [...] }` objects (via `StyleInputData` / `PropInputData`). Either the README or the API should be updated for consistency.

demo/story-template.ts
```diff
@@ -104,8 +107,15 @@ export class StoryTemplate extends LitElement {
...
+          .styleInputData=${this.styleInputData}
+          @stylesApplied=${this.handleStylesApplied}
+        ></story-styles-settings>
+        <story-props-settings
+          .propInputData=${this.propInputData}
+          @propsApplied=${this.handlePropsApplied}
+        ></story-props-settings>
+        ${when(this.shouldShowPropertySettings, () => html` <h3>Settings</h3>`)}
         <div
           class="slot-container"
```
bfalling, line 118: **Redundant conditional hiding:** `when(this.shouldShowPropertySettings, ...)` conditionally renders the `<h3>`, but the slot container below (line 121-126) is always rendered with an inline `style="display: none"` fallback. Consider wrapping both in the same conditional instead of mixing approaches.

demo/story-components/story-prop-settings.ts
```diff
@@ -0,0 +1,153 @@
...
+                  type="radio"
+                  class="prop-input"
+                  name=${inputId}
+                  id=${option}
+                  value=${option}
+                  data-prop=${settings.propertyName}
+                  ?checked=${settings.defaultValue === option}
+                /><label for=${option}> ${option} </label>`,
+          )}
```
bfalling, line 109: **Unconditional newline formatting:** The format `'\n  ' + stringifiedProps.join('\n  ') + '\n'` always adds leading/trailing newlines even for a single prop, which may look odd in the syntax highlighter for simple cases.

bfalling (commented):
Inline code review comments below. (via Claude)

demo/story-components/story-prop-settings.ts
rebecca-shoptaw: This is intentional; we want to allow for future expansion of this object to include more than just `settings`, as @jbuckner suggested on a previous MR. 

I'll update the README to match.

demo/story-template.ts
rebecca-shoptaw: Likely intentional, actually, fixed a previous issue with the CSS indentation so it is correct.

demo/story-template.ts
rebecca-shoptaw: This is necessary so that we can set props on the element via provided property settings without knowing the type of the element directly.

See:
```
appliedProps.forEach((prop) =>
      this.slottedDemoComponent?.setAttribute(prop.propName, prop.value),
    );
```

If the type is set to `Element` or `HTMLElement` this results in a type error and fails.

README.md
rebecca-shoptaw: README updated per above comment

demo/story-template.ts
rebecca-shoptaw: This approach is necessary because we want the slot to always be available in the DOM, so we can detect changes to it and update its visibility accordingly, but we don't want to render a hidden `Settings` title if it's not necessary so that we don't confuse screen readers.

demo/story-components/story-prop-settings.ts
rebecca-shoptaw: I don't find it odd, especially since we're usually setting multiple props, but that's more of a design decision for Jim S when the time comes

demo/story-components/story-prop-settings.ts
```diff
@@ -0,0 +1,153 @@
...
+      <tr>
+        <td><legend>${settings.label}</legend></td>
+        <td>
+          ${settings.radioOptions.map(
+            (option) =>
+              html`<input
+                  type="radio"
+                  class="prop-input"
+                  name=${inputId}
+                  id=${'${inputId}-${option}'}
```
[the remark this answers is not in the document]
rebecca-shoptaw: Since this is within a Lit template, the variable doesn't need backticks, but in this case the extra `${}` was unnecessary. Will remove.

demo/story-components/story-prop-settings.ts
```diff
@@ -0,0 +1,153 @@
...
+                  value=${option}
+                  data-prop=${settings.propertyName}
+                  ?checked=${settings.defaultValue === option}
+                /><label for=${'${inputId}-${option}'}> ${option} </label>`,
```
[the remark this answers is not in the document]
rebecca-shoptaw: See above

bfalling (approved):
LGTM

PR #37, opened 2026-03-12
Demo enhancements

pezvi:
- New 2 column layout
- Restyled element presentation
- Automatically add elements to the demo

### DIFF
--- a/demo/app-root.ts
+++ b/demo/app-root.ts
@@ -24,7 +24,7 @@ const ALL_ENTRIES       = [...productionEntries, ...labsEntries];
 export class AppRoot extends LitElement {
   createRenderRoot() { return this; }
 
-  private _scrollHandler?: () => void;
+  private _observer?: IntersectionObserver;
 
   render() {
     return html`
@@ -59,38 +59,41 @@ export class AppRoot extends LitElement {
       allIds.map(id => [id, this.querySelector(`#ia-sidebar a[href="#${id}"]`)])
     );
 
-    const getAnchorTops = () =>
-      allIds.map(id => {
-        const el = document.getElementById(id);
-        return el ? el.getBoundingClientRect().top + window.scrollY : null;
-      });
+    const visible = new Set<string>();
 
-    const updateActive = () => {
-      const tops = getAnchorTops();
-      let current = 0;
-      for (let i = 0; i < tops.length; i++) {
-        if ((tops[i] ?? Infinity) <= window.scrollY + 40) current = i;
-      }
-      allIds.forEach((id, i) => {
-        links[id]?.classList.toggle('active', i === current);
-      });
-    };
+    // Only anchors in the top 30% of the viewport count as "active".
+    // The first (topmost) visible anchor wins.
+    this._observer = new IntersectionObserver(
+      entries => {
+        for (const entry of entries) {
+          if (entry.isIntersecting) visible.add(entry.target.id);
+          else visible.delete(entry.target.id);
+        }
+        const activeId = allIds.find(id => visible.has(id)) ?? allIds[0];
+        allIds.forEach(id => links[id]?.classList.toggle('active', id === activeId));
+      },
+      { rootMargin: '0px 0px -70% 0px' },
+    );
 
-    allIds.forEach((id, i) => {
-      links[id]?.addEventListener('click', (e: Event) => {
-        e.preventDefault();
-        const top = getAnchorTops()[i];
-        if (top !== null) window.scrollTo({ top: Math.max(0, top - 16), behavior: 'smooth' });
-      });
+    allIds.forEach(id => {
+      const el = document.getElementById(id);
+      if (el) this._observer!.observe(el);
     });
 
-    this._scrollHandler = updateActive;
-    window.addEventListener('scroll', updateActive, { passive: true });
-    updateActive();
+    allIds.forEach(id => {
+      links[id]?.addEventListener('click', (e: Event) => {
+        e.preventDefault();
+        const el = document.getElementById(id);
+        if (el) {
+          const top = el.getBoundingClientRect().top + window.scrollY;
+          window.scrollTo({ top: Math.max(0, top - 16), behavior: 'smooth' });
+        }
+      });
+    });
   }
 
   disconnectedCallback() {
     super.disconnectedCallback();
-    if (this._scrollHandler) window.removeEventListener('scroll', this._scrollHandler);
+    this._observer?.disconnect();
   }
 }
--- a/demo/app-root.ts
+++ b/demo/app-root.ts
@@ -13,20 +13,22 @@ const storyEntries = Object.keys(storyModules)
   .map(path => {
     const labs = path.includes('/src/labs/');
     const parts = path.split('/');
-    const tag = parts[parts.length - 2];
+    const filename = parts[parts.length - 1]; // e.g. "ia-button-story.ts"
+    const tag = filename.replace(/-story\.ts$/, '');
     return { tag, storyTag: `${tag}-story`, id: `elem-${tag}`, labs };
   })
   .sort((a, b) => a.tag.localeCompare(b.tag));
 
 const productionEntries = storyEntries.filter(e => !e.labs);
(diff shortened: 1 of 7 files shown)


demo/app-root.ts
```diff
@@ -1,26 +1,96 @@
...
+import { unsafeHTML } from 'lit/directives/unsafe-html.js';
 
-import '@src/elements/ia-button/ia-button-story';
-import '@src/labs/ia-snow/ia-snow-story';
-import '@src/elements/ia-combo-box/ia-combo-box-story';
-import '@src/elements/ia-status-indicator/ia-status-indicator-story';
+const storyModules = import.meta.glob(
+  ['../src/elements/**/*-story.ts', '../src/labs/**/*-story.ts'],
+  { eager: true }
+);
```
jbuckner, line 10: Whoa! Nice!

demo/app-root.ts
```diff
@@ -1,26 +1,101 @@
...
+
+const storyEntries = Object.keys(storyModules)
+  .map(path => {
+    const labs = path.includes('/src/labs/');
+    const parts = path.split('/');
+    const tag = parts[parts.length - 2];
```
bfalling: Claude: This derivation assumes every story lives at `<element-name>/<element-name>-story.ts`. If a story file ever lives at a different nesting depth, `parts[parts.length - 2]` will silently produce the wrong tag. Consider extracting a named helper with a comment documenting the expected convention, or matching on the `-story.ts` filename itself (e.g. strip the `-story` suffix).

demo/app-root.ts
```diff
@@ -1,26 +1,101 @@
...
+    return { tag, storyTag: `${tag}-story`, id: `elem-${tag}`, labs };
+  })
+  .sort((a, b) => a.tag.localeCompare(b.tag));
+
+const productionEntries = storyEntries.filter(e => !e.labs);
+const labsEntries       = storyEntries.filter(e => e.labs);
+const ALL_ENTRIES       = [...productionEntries, ...labsEntries];
```
bfalling: Claude: Nit: the extra spaces for alignment (`labsEntries       `, `ALL_ENTRIES       `) will likely conflict with Prettier / your formatter. Alignment-based spacing tends to break as soon as a nearby variable name changes length.

demo/app-root.ts
```diff
@@ -1,26 +1,101 @@
...
+
+    allIds.forEach(id => {
+      links[id]?.addEventListener('click', (e: Event) => {
+        e.preventDefault();
+        const el = document.getElementById(id);
+        if (el) {
+          const top = el.getBoundingClientRect().top + window.scrollY;
+          window.scrollTo({ top: Math.max(0, top - 16), behavior: 'smooth' });
+        }
+      });
```
bfalling: Claude: The click listeners added here are never removed in `disconnectedCallback` — only the `IntersectionObserver` is cleaned up. For a singleton `app-root` this is fine in practice, but it's inconsistent with the observer cleanup pattern. If you wanted symmetry, you could use an `AbortController` and call `controller.abort()` in `disconnectedCallback`.

demo/index.css
```diff
@@ -24,7 +24,89 @@ a:hover {
...
+  color: #194880;
+  border-left-color: #194880;
+  background: #f0f4f9;
+  font-weight: 600;
+}
+
+#ia-content {
+  flex: 1;
+  padding: 0.75rem 1rem;
+  padding-bottom: 100vh;
```
bfalling: Claude: `padding-bottom: 100vh` is presumably so the scroll-spy can activate the last element in the sidebar (since it needs the content to scroll past the 30% viewport threshold). Worth a short comment explaining why — this looks like a bug at first glance.

demo/story-template.ts
```diff
@@ -69,72 +70,140 @@ export class StoryTemplate extends LitElement {
...
           ></slot>
         </div>
-        <h3>Import</h3>
+        <button
+          class="details-toggle ${this.detailsVisible ? '' : 'collapsed'}"
+          @click=${() => (this.detailsVisible = !this.detailsVisible)}
+        >
+          Import, Usage &amp; Settings
+        </button>
+        <div id="details" class="${this.detailsVisible ? '' : 'collapsed'}">
```
bfalling: Claude: Minor: `class="${this.detailsVisible ? '' : 'collapsed'}"` produces `class=""` in the DOM when visible. Consider using Lit's `classMap` directive for cleaner output:
```ts
class=${classMap({ collapsed: !this.detailsVisible })}
```

demo/story-template.ts
```diff
@@ -69,72 +70,140 @@ export class StoryTemplate extends LitElement {
...
+    text: string,
+    which: 'import' | 'usage' | 'styling',
+  ): Promise<void> {
+    try {
+      await navigator.clipboard.writeText(text);
+      this.copiedKey = which;
+      clearTimeout(this._copyTimeout);
+      this._copyTimeout = setTimeout(() => (this.copiedKey = null), 2000);
+    } catch {
+      // Clipboard API unavailable (non-HTTPS or permission denied) — silent fail
```
bfalling: Claude: The silent catch is intentional per the comment, but a `console.warn` would help when debugging clipboard issues on localhost (which _is_ where this demo page runs). Something like:
```ts
catch (e) {
  console.warn('Clipboard write failed:', e);
}
```
Still non-blocking, but visible in devtools.

demo/story-template.ts
```diff
@@ -69,72 +70,140 @@ export class StoryTemplate extends LitElement {
...
+              !!this.propInputData,
+              () => html`
+                <story-props-settings
+                  .propInputData=${this.propInputData}
+                  @propsApplied=${this.handlePropsApplied}
+                ></story-props-settings>
+              `,
+            )}
+            ${when(
+              !this.propInputData && !this.shouldShowPropertySettings,
```
bfalling, line 159: Claude: There may be a brief flash of "No settings to adjust" before the `slotchange` event fires, even when settings _are_ slotted. `shouldShowPropertySettings` starts `false` and only flips on `slotchange`, which fires asynchronously after first render. If this is noticeable, one option is to hide the placeholder until the first `slotchange` has been processed.

bfalling (commented):
Claude: Overall this is a nice improvement — the auto-discovery via `import.meta.glob`, sidebar nav with scroll-spy, collapsible details panel, and copy buttons are all solid additions. The test file is thorough. A few comments inline.

demo/app-root.ts
pezvi: Switched to using filename

demo/story-template.ts
pezvi: Now uses 'collapsed' and the (semantic) 'expanded' class string

demo/story-template.ts
pezvi: Flash is not noticeable

demo/app-root.ts
```diff
@@ -1,26 +1,104 @@
...
+    // Only anchors in the top 30% of the viewport count as "active".
+    // The first (topmost) visible anchor wins.
+    this._observer = new IntersectionObserver(
+      entries => {
+        for (const entry of entries) {
+          if (entry.isIntersecting) visible.add(entry.target.id);
+          else visible.delete(entry.target.id);
+        }
+        const activeId = allIds.find(id => visible.has(id)) ?? allIds[0];
+        allIds.forEach(id => links[id]?.classList.toggle('active', id === activeId));
```
rebecca-shoptaw, line 77: Can all of this logic be done inside a single `forEach` loop? Or two separate ones? It would also be great to have separate named functions for the logical pieces, i.e. `toggleActiveClasses(id)`, `addObserver(id)`, `scrollTo(id)` etc.

The repeated `getElementById` checks also feel redundant and un-Lit-ish; is there somewhere else we can store references to these elements? Or if not, it would be great to just do the check in one place

pezvi: Yes, I like that. Thanks, Rebecca!

jbuckner (approved):
I think @rebecca-shoptaw and @bfalling picked up on all of the things. Looks good! 👍

PR #39, opened 2026-04-24
Elements: Move ia-zendesk-widget component into elements repo

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-8356


src/elements/ia-zendesk-widget/ia-zendesk-widget-story.ts
```diff
@@ -0,0 +1,76 @@
...
+import type { IAZendeskWidget } from './ia-zendesk-widget';
+
+import './ia-zendesk-widget';
+import '@demo/story-template';
+
+const WIDGET_SRC_TEST =
+  'https://static.zdassets.com/ekr/snippet.js?key=6fe87bd8-d4e3-4b42-8632-be6eb933d54d';
+
+const WIDGET_SRC_PROD =
+  'https://static.zdassets.com/ekr/snippet.js?key=685f6dc4-48c5-411f-8463-cc6dd50abe2d';
```
jbuckner: For the demo, I don't think we need to include the prod key. This should just be provided by the host app.

src/elements/ia-zendesk-widget/ia-zendesk-widget.ts
```diff
@@ -0,0 +1,246 @@
...
+  css,
+  html,
+  LitElement,
+  type CSSResultGroup,
+  type TemplateResult,
+} from 'lit';
+import { customElement, property, state } from 'lit/decorators.js';
+import { unsafeHTML, UnsafeHTMLDirective } from 'lit/directives/unsafe-html.js';
+
+import loaderIconSvg from './loader-icon.svg?raw';
```
jbuckner: How come you're using `?raw`?

src/elements/ia-zendesk-widget/ia-zendesk-widget.ts
```diff
@@ -0,0 +1,246 @@
...
+    zE?: {
+      (
+        target: 'messenger:on',
+        event: 'open' | 'close',
+        callback: () => void,
+      ): void;
+      (target: 'messenger', action: 'open' | 'close'): void;
+    };
+  }
+}
```
jbuckner: Can we pull this out to its own file?

src/elements/ia-zendesk-widget/ia-zendesk-widget.ts
```diff
@@ -0,0 +1,246 @@
...
+    }
+  }
+
+  /**
+   * Appends the Zendesk snippet `<script>` to `<head>` if it has not already
+   * been added. Resolves once the script has loaded; rejects on network error.
+   * Guarded by the `ze-snippet` id so multiple widget instances on the same
+   * page share a single script tag.
+   */
+  private loadZendeskScript(): Promise<void> {
```
jbuckner: Can we use async/await instead of Promises?

jbuckner: Oh wait I see the `onload` / `onerror` uses the resolve/reject

src/elements/ia-zendesk-widget/ia-zendesk-widget.ts
```diff
@@ -0,0 +1,246 @@
...
+        clearInterval(check);
+        reject(new Error('Zendesk API did not initialise in time'));
+      }, timeoutMs);
+    });
+  }
+
+  private get iconTemplate(): DirectiveResult<typeof UnsafeHTMLDirective> {
+    return this.isLoading
+      ? unsafeHTML(loaderIconSvg)
+      : unsafeHTML(questionIconSvg);
```
jbuckner: You should be able to return the actual svg without using unsafeHTML

src/elements/ia-zendesk-widget/ia-zendesk-widget.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    try {
+      if (!this.zendeskReady) {
+        await this.loadZendeskScript();
+        await this.waitForZendesk();
+
+        // Register lifecycle listeners exactly once.
+        // isLoading is cleared here (not earlier) so the spinner persists until
+        // the widget panel is actually visible to the user.
+        window.zE!('messenger:on', 'open', () => {
```
jbuckner: There are several `ze!()` here. Instead of force-unwrapping, we should probably do something like `if (window.zE) { ... }` or `if (!window.zE) { return; }`

src/elements/ia-zendesk-widget/ia-zendesk-widget.ts
nsharma123: Its added by Claude. It imports the SVG file as a plain string instead of a URL. That gives us the raw <svg>...</svg> markup, which we then render using unsafeHTML. Now I changed this.

PR #40, opened 2026-04-29
WEBDEV-8327 Add basic OTP input

rebecca-shoptaw:
Builds out a simple OTP input component (just the boxes and pasting/arrow key behavior), migrated from an existing Offshoot component.

Also expands the prop settings to support boolean and number props.

### DIFF
--- a/demo/story-components/story-prop-settings.ts
+++ b/demo/story-components/story-prop-settings.ts
@@ -132,7 +132,7 @@ export class StoryPropsSettings extends LitElement {
       if (value === 'false') value = false;
 
       const stringifiedValue =
-        typeof value === 'string' ? `'${value}` : value.toString();
+        typeof value === 'string' ? `'${value}'` : value.toString();
       stringifiedProps.push(`.${propName}=\${${stringifiedValue}}`);
       appliedProps.push({ propName, value });
     });
--- a/demo/story-components/story-prop-settings.ts
+++ b/demo/story-components/story-prop-settings.ts
@@ -9,7 +9,6 @@ import {
 import { property, queryAll } from 'lit/decorators.js';
 import { customElement } from 'lit/decorators/custom-element.js';
 import { choose } from 'lit/directives/choose.js';
-import { ifDefined } from 'lit/directives/if-defined.js';
 
 import themeStyles from '@src/themes/theme-styles';
 import { labelToId } from '../story-utils';
@@ -17,7 +16,7 @@ import { labelToId } from '../story-utils';
 export type PropInputSettings<T> = {
   label: string;
   propertyName: keyof T;
-  defaultValue?: string | boolean | number;
+  defaultValue: string | boolean | number;
   inputType?: 'text' | 'radio' | 'number';
   radioOptions?: string[] | boolean[];
 };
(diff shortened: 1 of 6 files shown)


demo/story-components/story-prop-settings.ts
```diff
@@ -124,8 +124,17 @@ export class StoryPropsSettings extends LitElement {
...
 
       const propName = input.dataset.prop;
-      stringifiedProps.push(`.${propName}=\${'${input.value}'}`);
-      appliedProps.push({ propName, value: input.value });
+      let value: number | string | boolean = input.value;
+
+      // Perform necessary conversions for props to apply
+      if (input.type === 'number') value = parseInt(value);
+      if (value === 'true') value = true;
+      if (value === 'false') value = false;
```
latonv: Nice extension of the property types!

Q: Might these conversions be overly broad? In the current example, this means that a (non-numeric) OTP prefill value of the string `true` ends up trying incorrectly to apply the boolean, for instance.

demo/story-components/story-prop-settings.ts
```diff
@@ -124,8 +124,17 @@ export class StoryPropsSettings extends LitElement {
...
+
+      const stringifiedValue =
+        typeof value === 'string' ? `'${value}` : value.toString();
```
latonv: Looks like an extra apostrophe may have snuck into `'${value}` here

src/elements/ia-otp-input/ia-otp-input.ts
```diff
@@ -0,0 +1,234 @@
...
+const NUMERIC_REGEX = /^[0-9]+$/;
+const ALPHANUMERIC_REGEX = /^[a-zA-Z0-9]+$/;
+
+/**
+ * Renders an input for OTP entry, which combines a series of separate
+ * inputs into one, shifting focus as needed and handling paste and backspace
+ * events.
+ */
+@customElement('ia-otp-input')
+export class IAOTPInput extends LitElement {
```
latonv, line 28: Nicely written component, very easy to understand what's going on!

src/elements/ia-otp-input/ia-otp-input.ts
```diff
@@ -0,0 +1,234 @@
...
+    const next = input.nextElementSibling as HTMLInputElement;
+    if (next) next.focus();
+
+    this.submitIfInputsFilled();
+  }
+
+  /* Performs special actions associated with some keys */
+  private handleKeydown(e: KeyboardEvent): void {
+    const input = e.target as HTMLInputElement;
+    const key = e.key.toLowerCase();
```
latonv: Q: Just curious in passing why we need to lowercase the key codes here. Is it for wider browser support?

latonv (approved):
Looks great! Mostly just noted a couple of small nitpicks in the story settings.

demo/story-components/story-prop-settings.ts
rebecca-shoptaw: You're right, I think I'll try storing the type info on the input so we can check that directly and avoid inferring incorrectly like this.

demo/story-components/story-prop-settings.ts
rebecca-shoptaw: Actually one is missing! We want to put it in quotes in the demo (if it's a string).

src/elements/ia-otp-input/ia-otp-input.ts
rebecca-shoptaw: I'm not sure actually, it's something I've done out of habit presumably because I read something about it somewhere, but I'm not seeing the rationale for it when I look now, so I'll experiment with taking it out.

PR #41, opened 2026-05-01
WEBDEV-8378 Migrate dropdown-style search bar from Offshoot

latonv:
We currently have a search bar component in Offshoot that incorporates a built-in dropdown menu with search categories. We'd like to use this in contexts outside of Offshoot too, so we are migrating it into Elements.

The component has a custom `categories` property defining what options to show in the dropdown, as well several further options like whether/how to show an Advanced Search link and whether to show its loading state. One may also omit the dropdown piece entirely to use it as a plain search bar if desired.

Notably, this component presently depends on two other packages `@internetarchive/ia-dropdown` and `@internetarchive/ia-clearable-text-input` which are slated to eventually be migrated here as well.


src/elements/ia-dropdown-search-bar/ia-dropdown-search-bar-story.ts
```diff
@@ -0,0 +1,243 @@
...
+      advancedSearchStyle: advancedSearchStyle ? `"${advancedSearchStyle}"` : '',
+      hideDropdown: this.hideDropdown,
+      loading: this.loading,
+    };
+
+    const bindingsStr = Object.entries(bindings)
+      .map(([key, val]) => {
+        if (!val) return '';
+        if (val === true) return key;
+        return val ? `${key}=${val}` : '';
```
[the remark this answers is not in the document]
latonv: Fixed in https://github.com/internetarchive/elements/pull/41/commits/ba9c07c7464418d3109717bb7c209f87165cbabd

src/elements/ia-dropdown-search-bar/ia-dropdown-search-bar-story.ts
```diff
@@ -0,0 +1,304 @@
...
+    return html`
+      <story-template
+        elementTag="ia-dropdown-search-bar"
+        elementClassName="IADropdownSearchBar"
+        .customExampleUsage=${this.exampleUsage}
+        .styleInputData=${{ settings: styleInputSettings }}
+      >
+        <div slot="demo">
+          <ia-dropdown-search-bar
+            .navBaseUrl=${''}
```
jbuckner: Should we default this to https://archive.org for the demo? When I test it, it goes to https://internetarchive.github.io/advancedsearch.php so it 404s

latonv: That would make sense!

In fact, I'll be removing all of the Advanced Search stuff from this component so that it generalizes better, so we may not actually need this nav URL anymore to begin with.

jbuckner (approved):

PR #45, opened 2026-05-13
WEBDEV-8447 Upgrade ia-dropdown version to address z-index bug

latonv:
There is currently a bug with incorrect z-index behavior on the `ia-dropdown`. This upgrades to a newer version of that component that fixes the bug, to avoid component registry conflicts when the `ia-dropdown-search-bar` is used alongside other components that rely on the patched version of `ia-dropdown`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   "dependencies": {
     "@lit/localize": "^0.12.2",
     "@internetarchive/ia-clearable-text-input": "^1.1.1",
-    "@internetarchive/ia-dropdown": "^2.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
   "dependencies": {
     "@lit/localize": "^0.12.2",
     "@internetarchive/ia-clearable-text-input": "^1.1.1",
-    "@internetarchive/ia-dropdown": "^2.1.0",
+    "@internetarchive/ia-dropdown": "2.2.0-alpha-webdev8447.0",
```
latonv: (To be upgraded to an equivalent production version before merging)

bfalling (approved):
LGTM

PR #46, opened 2026-05-19
WEBDEV-8464 Make ia-dropdown-search-bar propagate selection changes immediately

latonv:
Currently for `<ia-dropdown-search-bar>`, if one resets the `selectedCategory` property directly after a dropdown option is selected, the change does not propagate down to the inner ia-dropdown component via Lit's usual update cycle due to the property-binding diff seeing no change. This PR just propagates the new value immediately via `willUpdate` to ensure the same change is seen by every component in the tree.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@internetarchive/elements",
-  "version": "0.2.6",
+  "version": "0.2.7",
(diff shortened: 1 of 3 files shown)


rebecca-shoptaw (approved):
Nice workaround, and helpful explanation!

PR #50, opened 2026-05-27
WEBDEV-8506: Add conventions for non-component modules

jbuckner:
Establishes the directory layout + `package.json` exports + test conventions for non-component modules (`src/services`, `src/parsers`, `src/models`, `src/types`) ahead of consolidating `metadata-service` and its deps into elements. Docs + routing rules only — no code moved.

- `README.md`: "Adding a Service or Library Module" section (dirs, structure, exports, tests).
- `package.json`: subpath exports for the four new dirs (mirrors `./labs/*`).

Lands conventions first so the four migration PRs (WEBDEV-8507/8508/8509/8510) are straight code-moves. Build green.

[WEBDEV-8506](https://webarchive.jira.com/browse/WEBDEV-8506)

[WEBDEV-8506]: https://webarchive.jira.com/browse/WEBDEV-8506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #51, opened 2026-05-27
WEBDEV-8507: Migrate field-parsers into elements

jbuckner:
Move `@internetarchive/field-parsers` source + tests into `src/parsers/` per the WEBDEV-8506 conventions ([#50](https://github.com/internetarchive/elements/pull/50)). First step consolidating the `metadata-service` dependency tree into elements.

- `field-parser-interface.ts` + 10 parsers — straight copy.
- `timed.ts` / `trace.ts` debug decorators — brought for parity (retire later if unused).
- Tests colocated, ported Web Test Runner/Mocha/Chai → Vitest (mechanical: `it`→`test`, `.to.be.true`→`toBe(true)`, `.to.deep.equal`→`toEqual`).
- `epoch.ts`: `enum`→`const … as const` for elements' `erasableSyntaxOnly`.

No runtime change; build + tests green. QA-waived (code move).

[WEBDEV-8507](https://webarchive.jira.com/browse/WEBDEV-8507)

[WEBDEV-8507]: https://webarchive.jira.com/browse/WEBDEV-8507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/parsers/field-types/date.ts
```diff
@@ -0,0 +1,57 @@
...
+  // instantiating a new instance for every use
+  static shared = new DateParser();
+
+  /** @inheritdoc */
+  parseValue(rawValue: FieldParserRawValue): Date | undefined {
+    // try different date parsing
+    return this.parseJSDate(rawValue) || this.parseBracketDate(rawValue);
+  }
+
+  // handles "[yyyy]" format
```
bfalling, line 17: QUESTION: Are there any month/year formats we need to support? (BTW: @ximm is a good reference for dates we support throughout the system.)

src/parsers/field-types/epoch.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+  }
+
+  @timed()
+  @trace2
+  someSlowMethod() {
+    for (let i = 0; i < 1e6; i++) {
+      // EpochParser.secondsParser.parseValue(1637274397); // Example usage of the parser
+    }
+  }
```
bfalling: QUESTION: What is this method doing here?

src/parsers/field-types/list.ts
```diff
@@ -0,0 +1,45 @@
+import {
+  FieldParserInterface,
+  FieldParserRawValue,
+} from '../field-parser-interface';
+
+export class ListParser<T> implements FieldParserInterface<T[]> {
```
bfalling, line 6: SUGGESTION: It'd be nice if each parser data type listed some example values that it might receive in the comments at the top of the file.

src/parsers/field-types/mediatype.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    expect(parser.parseValue('account')).toBe('account');
+    expect(parser.parseValue('audio')).toBe('audio');
+    expect(parser.parseValue('collection')).toBe('collection');
+    expect(parser.parseValue('data')).toBe('data');
+    expect(parser.parseValue('etree')).toBe('etree');
+    expect(parser.parseValue('image')).toBe('image');
+    expect(parser.parseValue('movies')).toBe('movies');
+    expect(parser.parseValue('software')).toBe('software');
+    expect(parser.parseValue('texts')).toBe('texts');
+    expect(parser.parseValue('web')).toBe('web');
```
bfalling, line 17: QUESTION: Is this actually a useful set of tests? What if one passed in a value that wasn't in the `MediaType` set?

src/parsers/timed.ts
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export function timed(): (
```
bfalling: SUGGESTION: Can we be a bit more verbose? I'd think that throughout the code, any non-trivial functions should have doc blocks, explaining what they're for and how to use them. I see a lot fewer in this MR than usual. Anyway, I'd love to know more about the purpose and usage of this function.

src/parsers/field-types/date.ts
jbuckner: Potentially? This PR is just for migrating from a standalone repo into elements, not modifying any functionality. I filed https://webarchive.jira.com/browse/WEBDEV-8535 to investigate that.

ximm: > QUESTION: Are there any month/year formats we need to support? (BTW: @ximm is a good reference for dates we support throughout the system.)

I can talk about dates, but would need context. 

(In brief, forgiving raw string formats with enormous variability and many styles are explicitly allowed in the `date` field in item metadata; these are normalized in search documents into ISO8601 timestamps, and `year` derived when possible. Related, I just merged a chance today that will extend the representation of dates in all search documents, to exprsss a range, with a best-effort precision and uncertainty assignment.)

PR #60, opened 2026-06-24
WEBDEV-8504 Add ia-sr-only-text to Elements

rebecca-shoptaw:
Adds a nice helper element which makes any text slotted into it visible only to screen readers, but not on the page.


pezvi (approved):
LGTM

PR #65, opened 2026-07-22
WEBDEV-8777 Apply ia-button custom active styles for .active class

nsharma123:
The custom variant's active-state selector only matched the native :hover/:focus/:active pseudo-classes, so elements that toggle a persistent .active class via JS never got the active styling.

https://webarchive.jira.com/browse/WEBDEV-8777


jbuckner (approved):

pezvi (approved):
lgtm

PR #71, opened 2026-08-06
WEBDEV-8711 add type check and value conversion at the top of the function

dualcnhq:


src/elements/ia-otp-input/ia-otp-input.ts
```diff
@@ -155,6 +155,8 @@ export class IAOTPInput extends LitElement {
    * Submits the result if enough characters are filled.
    */
   private fillInputs(value: string) {
+    if (typeof value !== 'string') value = String(value ?? '');
```
jbuckner, line 158: **This coerces anything, not just numbers.** `numericOnly` defaults to `true` so the regex drops the garbage and we're fine in practice, but at `numericOnly=false` an object prefill becomes `"[object Object]"` and fills the boxes with `objectO`. Bailing reads safer than coercing for anything that isn't a string or a number:

```ts
if (typeof value === 'number') value = String(value);
if (typeof value !== 'string') return;
```

Also `String(value ?? '')` turns a `null` prefill into `''`, which then falls into `clearInputs()`. Probably what you want, just flagging it's a behavior change.

Q: was the number case the actual intent here, or just the easiest thing to test?

src/elements/ia-otp-input/ia-otp-input.test.ts
```diff
@@ -575,6 +575,21 @@ describe('IA OTP Input', () => {
     expect(inputs?.[5].value).to.equal('');
   });
 
+  test('does not throw if prefill value property is set to a non-string value', async () => {
```
jbuckner, line 578: **This covers the one path that never crashed.** The reported error is `handleInput` reading `e.data` off `beforeinput`, not `prefillValue`. If we're hardening against a non-string reaching `fillInputs`, worth firing a `beforeinput` with a non-string `data` too:

```ts
const ev = new InputEvent('beforeinput', { bubbles: true });
Object.defineProperty(ev, 'data', { value: {} });
inputs?.[0].dispatchEvent(ev);
```

jbuckner:
Left two inline notes.

Bigger picture though: I don't think this error is ours. Wrote it up on [WEBDEV-8711](https://webarchive.jira.com/browse/WEBDEV-8711) with the detail. Short version: `fillInputs` minifies to `e.split(...)` in the live offshoot bundle so it can't produce `n.data.split`, there's no `.data.split(` anywhere in our output, and the Sentry group it was filed from is a bucket of frameless browser-extension errors.

Happy to take this as hardening on its own merits, I just don't want it closing the ticket.


[WEBDEV-8711]: https://webarchive.jira.com/browse/WEBDEV-8711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #76, opened 2026-08-20
WEBDEV-8927: Add a focus mode to the elements demo

jbuckner:
The demo showed every element on one page. Now `#elem-<tag>` shows just that one, the sidebar switches the focused element instead of scrolling to it, and "All elements" goes back to the full list. Story modules load lazily, so a deep link only fetches that element's chunk.

Hash routing because the demo ships to GH Pages under a base path it doesn't know at build time, so a real path would 404 on reload.

WEBDEV-8927: https://webarchive.jira.com/browse/WEBDEV-8927

https://claude.ai/code/session_01QEWomUzGqJ8SazBDMrZEti

### DIFF
--- a/demo/index.css
+++ b/demo/index.css
@@ -72,14 +72,6 @@ app-root {
   color: #213547;
 }
 
-/* Filled rather than the tinted treatment the element links use, so the view
-   you're in stays distinct from the scroll position within it. */
-#ia-all-link.active {
-  background: #194880;
-  border-color: #194880;
-  color: #fff;
-}
-
 #ia-sidebar h2 {
   font-size: 0.7rem;
   font-weight: 600;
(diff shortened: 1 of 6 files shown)


demo/index.css
```diff
@@ -46,6 +46,15 @@ app-root {
   overflow-y: auto;
 }
 
+#ia-sidebar #ia-all-link {
```
latonv, line 49: Is there a need for both of these ids in the selector? Would just the final one suffice?

latonv (approved):
LGTM, just noted a CSS selector that could possibly be simplified if desired.

demo/index.css
jbuckner: Good catch, though the answer surprised me: just `#ia-all-link` would not have worked. `#ia-sidebar a` is (1,0,1) and a bare `#ia-all-link` is (1,0,0), so the generic link rule below it would win on color, font-weight and padding. The second id was only there to beat it.

So I got rid of the overlap instead. The element links now carry an `.ia-elem-link` class, the generic rule is scoped to that, and `#ia-all-link` stands on its own like you wanted.

Jason also had some feedback on this, so the same push restyles that link as an actual button labeled "Show All Elements", highlights it when the all-elements view is showing, and adds the missing gap above "Labs".

PR #77, opened 2026-08-26
WEBDEV-8963: Migrate audio-element into elements

jbuckner:
[WEBDEV-8963](https://webarchive.jira.com/browse/WEBDEV-8963). Bottom of the radio player stack ([WEBDEV-8962](https://webarchive.jira.com/browse/WEBDEV-8962)), the other six components go on top of this one.

Two things that aren't just a file move:

- `vite.config.ts` passes `--autoplay-policy=no-user-gesture-required` to Playwright. Karma set this itself and the Vitest provider doesn't, so without it every playback test fails.
- The element re-dispatches a media `error` now. It never did, and since `play()` has to catch its own rejection for Vitest, there was otherwise no way to tell a dead source from silence. It only fires once the browser runs out of `<source>` candidates, so a failing source with a working fallback behind it stays quiet.

`AudioSource` is an interface instead of a class. It has no behaviour and the old tests already passed it as plain JSON.

[WEBDEV-8963]: https://webarchive.jira.com/browse/WEBDEV-8963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


latonv (approved):
Nice migration, LGTM!

PR #78, opened 2026-08-26
WEBDEV-8964: Migrate waveform-progress into elements

jbuckner:
[WEBDEV-8964](https://webarchive.jira.com/browse/WEBDEV-8964). Stacked on [#77](https://github.com/internetarchive/elements/pull/77), part of [WEBDEV-8962](https://webarchive.jira.com/browse/WEBDEV-8962).

Not just a file move. Five fixes came with it, none of which the old tests could catch:

- Scrub position comes from `getBoundingClientRect()` instead of `container.offsetLeft`. `offsetLeft` is measured from the offsetParent and `pageX` from the document, so they only agreed when the element sat at the page origin with nothing positioned above it. The old test hardcoded the off-by-the-body-margin result as its expected value, so it passed *because* of the bug. New tests derive the expectation from the element's own measured rect.
- Touch drags work now. The touch handlers were wired to a mouse-only handler reading `e.pageX`, which is `undefined` on a `TouchEvent`, so every touch drag computed `NaN`.
- `touchcancel` ends a drag. Only `touchend` did, so a touch the system took over (incoming call, browser gesture) left the drag latched and the waveform stopped following playback permanently.
- A non-primary `mousedown` is ignored. Right-clicking the waveform used to seek, and the context menu eats the `mouseup`, so it left the drag latched too.
- The emitted percentage is clamped to 0-100.

`ZoneOfSilence` is an interface, same reasoning as `AudioSource` in #77. CSS custom properties renamed to `--ia-theme-waveform-*`.

[WEBDEV-8964]: https://webarchive.jira.com/browse/WEBDEV-8964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbuckner:
Brenton, WEBDEV-8964 is signed off on your side already (brenton-cr-passed and brenton-qa-passed), but GitHub never got an approval on the PR itself. It's second from the bottom in the radio player stack, and stacks merge all-or-nothing, so this one is holding up the other seven. Mind clicking Approve?

Heads up that I merged main in just now to clear a conflict in src/elements/index.ts, so the diff moved a little since you looked. Lint is clean and all 615 tests pass on the top of the stack.

bfalling (approved):
LGTM

PR #80, opened 2026-08-27
WEBDEV-8965: Migrate playback-controls into elements

jbuckner:
[WEBDEV-8965](https://webarchive.jira.com/browse/WEBDEV-8965). Stacked on [#78](https://github.com/internetarchive/elements/pull/78), part of [WEBDEV-8962](https://webarchive.jira.com/browse/WEBDEV-8962).

Beyond the move:

- `PlaybackMode` was a numeric `enum`, which `erasableSyntaxOnly` rules out. It's a const object with string values now, so `playbackmode="playing"` works as an attribute. It never did: the component declared `@property({ type: PlaybackMode })`, which isn't a Lit converter, so the attribute path fell back to `String` and could only produce a string where the code compared against `0`/`1`.
- All seven buttons have accessible names. None had one before.
- Icons use `currentColor` instead of hardcoded `#fff`/`#333`, so the controls can sit on a light background. The default variable values reproduce the old colours exactly.
- Speed and volume clamp as well as wrap. An off-boundary value set from outside used to overshoot the range, and the `===` comparisons meant it could climb forever without ever hitting the wrap.
- `.jump-btn:active` targets the button rather than an `img`. The icons are inline SVG, so the old rule matched nothing and the jump buttons had no pressed state.

The speed button's label stops at the number rather than saying "1 times" — `@lit/localize` has no plural support, so a unit baked into the string can't be corrected per language.

Event names are unchanged, mixed kebab-case and camelCase and all, because `radio-player` listens for them later in this stack.

Note: this PR shows only `deploy-preview` because CI skips PRs not based on `main` ([WEBDEV-8972](https://webarchive.jira.com/browse/WEBDEV-8972)). Verified locally: `pnpm run build`, full suite 248 passing, `eslint` and `prettier` clean, 100% coverage on this component.

[WEBDEV-8965]: https://webarchive.jira.com/browse/WEBDEV-8965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/elements/ia-playback-controls/ia-playback-controls.ts
```diff
@@ -0,0 +1,338 @@
...
+      }
+
+      #play-pause-btn {
+        border-radius: 50%;
+        height: var(--playback-controls-play-button-diameter--);
+        width: var(--playback-controls-play-button-diameter--);
+        border: none;
+        background-color: var(--playback-controls-play-button-color--);
+        color: var(--playback-controls-play-icon-color--);
+        vertical-align: middle;
```
nsharma123, line 328: **The play icon doesn't scale proportionally with the new diameter variable.** This rule never resets the UA's default `padding: 1px 6px` on the button, and the SVG below is sized `width/height: 100%` of the *content* box. Buttons are `box-sizing: border-box` per the UA stylesheet, so the circle stays exactly the diameter but the icon's box loses 12px horizontally at every size.

Measured glyph-to-circle ratio in Chrome:

| `--ia-theme-playback-controls-play-button-diameter` | circle | icon box | glyph | % of circle |
|---|---|---|---|---|
| `2rem` | 32px | 20×30 | 20px | **63%** |
| `3rem` | 48px | 36×46 | 36px | 75% |
| `4rem` (default) | 64px | 52×62 | 52px | 81% |
| `8rem` | 128px | 116×126 | 116px | **91%** |

So the variable changes the icon's *proportions*, not just its size — the triangle is noticeably inset at small diameters and nearly touches the rim at large ones. (The `viewBox` is square with the default `preserveAspectRatio`, so it letterboxes rather than squashing — the glyph is just smaller than intended.)

One line fixes it:

```css
#play-pause-btn {
  border-radius: 50%;
  padding: 0;
  height: var(--playback-controls-play-button-diameter--);
  ...
```

The `.unstyled-button` reset below handles this for the other six buttons, but this one isn't in that class. This is pre-existing behaviour from iaux, but since this PR is what promotes the diameter to a supported theming knob, it seems like the right place to make it behave linearly.

src/elements/ia-playback-controls/ia-playback-controls.ts
```diff
@@ -0,0 +1,338 @@
...
+   *
+   * The comparison is `>=` rather than `===`, and the step is clamped, so a
+   * rate set from outside that doesn't land on a step boundary still stays
+   * inside the range instead of overshooting it or climbing forever.
+   */
+  private handlePlaybackRateChange(): void {
+    this.playbackRate =
+      this.playbackRate >= MAX_PLAYBACK_RATE
+        ? MIN_PLAYBACK_RATE
+        : Math.min(this.playbackRate + PLAYBACK_RATE_STEP, MAX_PLAYBACK_RATE);
```
nsharma123: **Only the upper bound is enforced — `MIN_PLAYBACK_RATE` is the wrap target but never a floor.** The description says "Both clamp and wrap now", and the wrap half is genuinely fixed (I confirmed `playbackRate="3.1"` wraps rather than climbing), but a value below the minimum still drifts permanently off the 0.25 grid.

Set `playbackRate = 0.1` and press the speed button:

```
0.1 → 0.35 → 0.6 → 0.85 → 1.1 → …
```

It never lands on any of the documented `0.5 / 0.75 / 1 / 1.25 / …` steps, and the label reads `0.85x` while the button announces "Playback speed, currently 0.85". It does eventually wrap once it passes 2, so this isn't the "climbs forever" bug — it's the other half of the same asymmetry.

Clamping the input as well as the output closes it:

```ts
const current = Math.max(this.playbackRate, MIN_PLAYBACK_RATE);
this.playbackRate =
  current >= MAX_PLAYBACK_RATE
    ? MIN_PLAYBACK_RATE
    : Math.min(current + PLAYBACK_RATE_STEP, MAX_PLAYBACK_RATE);
```

src/elements/ia-playback-controls/ia-playback-controls.ts
```diff
@@ -0,0 +1,338 @@
...
+    this.dispatchEvent(
+      new CustomEvent<{ playbackRate: number }>(Events.PlaybackRateChange, {
+        detail: { playbackRate: this.playbackRate },
+      }),
+    );
+  }
+
+  /** Steps the volume up, wrapping round to muted once it passes full. */
+  private handleVolumeChange(): void {
+    this.volume = this.volume >= 1 ? 0 : Math.min(this.volume + VOLUME_STEP, 1);
```
nsharma123: **Same missing lower bound here, and this one reaches the screen reader.** With `volume = -1`, pressing the volume button steps through negative values that are rendered and announced verbatim:

| press | `volume` | label | accessible name |
|---|---|---|---|
| 1 | -0.75 | `-75%` | "Volume, currently **-75** percent" |
| 2 | -0.5 | `-50%` | "Volume, currently -50 percent" |
| 3 | -0.25 | `-25%` | "Volume, currently -25 percent" |
| 4 | 0 | `0%` | "Volume, currently 0 percent" |

The same unguarded path also means a junk attribute survives into the a11y tree — `volume="loud"` gives `Number('loud')` → `NaN`, which renders `NaN%` and announces "Volume, currently NaN percent" (and picks the medium icon, since `NaN <= 0` and `NaN >= 1` are both false).

Clamping on the way in handles both:

```ts
const current = Math.min(Math.max(this.volume || 0, 0), 1);
this.volume = current >= 1 ? 0 : Math.min(current + VOLUME_STEP, 1);
```

Or clamp in the `volumePercent` / `volumeButtonIcon` getters if you'd rather keep the property as the consumer set it. Either way the accessible name stops reporting impossible values.

src/elements/ia-playback-controls/ia-playback-controls.test.ts
```diff
@@ -0,0 +1,252 @@
...
+
+    // The partial icon is the one with the faded outer arc.
+    el.volume = 0.5;
+    await elementUpdated(el);
+    expect(
+      el.shadowRoot?.querySelector(
+        '#volume-control-btn svg path[stroke-opacity]',
+      ),
+    ).to.exist;
+  });
```
nsharma123, line 288: **Nothing here covers the `color: inherit` fix, which is the subtlest of the six.** I deleted that line from `.unstyled-button` and re-ran this file: **all 26 tests still pass**, while every icon in the browser goes black. That's exactly the regression the fix exists to prevent, and it's also what makes the two icon-colour theme variables reach the icons at all — so a silent revert would take the theming with it.

A computed-style assertion locks it in, alongside these icon tests:

```ts
test('passes the icon colour theme variable down to the buttons', async () => {
  const el = await controlsFixture();
  el.style.setProperty('--ia-theme-playback-controls-icon-color', 'rgb(255, 0, 0)');
  await elementUpdated(el);

  expect(getComputedStyle(buttonIn(el, 'back-btn')).color).to.equal('rgb(255, 0, 0)');
  // The play button keeps its own colour rather than inheriting.
  expect(getComputedStyle(buttonIn(el, 'play-pause-btn')).color).to.equal('rgb(51, 51, 51)');
});
```

The `.jump-btn:active` opacity rule is in the same boat — I verified it by hand (mouse down → `opacity: 0.75`, release → `1`), but nothing would catch it going back to a selector that matches nothing, which is the bug it replaced. Worth one assertion if it's cheap to force `:active` in vitest browser mode; not worth contorting the test if it isn't.

nsharma123 (commented):
Reviewed the diff and ran the migrated component in the browser against every QA step in WEBDEV-8965. **All ten QA steps pass, and all six claimed bug fixes are real** — I confirmed the `:active` dim by holding the mouse down and reading `opacity: 0.75` (the old `:active img` rule genuinely matched nothing), and confirmed the theme variables reach both the icons and the `1x`/`100%` labels while leaving the play button untouched.

Diffed all ten assets against `iaux/packages/playback-controls` — 1:1 apart from `#FFF`/`#333` → `currentColor`. `--playPauseDiameter` has zero consumers anywhere in iaux/petabox/offshoot, so that rename is safe, and `radio-player` sets `playbackMode` via the imported `PlaybackMode`, so the numeric-enum → string change is source-compatible for the follow-up.

Build, prettier and eslint are clean (the 11 `no-explicit-any` warnings are pre-existing and all in `demo/`). Suite: 248 passed, 1 failed — `demo/app-root.test.ts`, which **also fails on the base branch** and passes in isolation, so it's a timing-sensitive demo test, not from this PR.

Nothing blocking. Two things worth fixing before merge (the play-button padding and the missing lower-bound clamp), plus some smaller notes inline.

One repo-wide note not worth anchoring: the assets use lit's `svg` tag on templates whose root is a complete `<svg>` element. It renders correctly and I verified all ten do, but `svg` is documented for SVG *fragments* — `html` is the conventional tag for a whole `<svg>`, and it's what the rest of the repo uses when inlining SVG. Cosmetic, take it or leave it.

Things I checked and deliberately did **not** flag: the private non-exported `const Events = {…}` without `as const` (exactly the convention in all five sibling elements), the missing `type="button"` (split in this repo, and harmless inside shadow DOM), and the events not bubbling / `playbackMode` not reflecting (both identical to the original).

src/elements/ia-playback-controls/assets/replay.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * @file Skip back ten seconds icon
+ *
+ * Inline so the button can recolor it with `currentColor`, which is not
+ * possible on the contents of an `<img>`.
+ */
+import { svg } from 'lit';
+
+export default svg`
+<svg width="20" height="20" viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg">
```
nsharma123, line 10: **Decorative icons could use `aria-hidden="true"`.** This icon and `skip-ahead.ts` contain a `<text>10</text>`, which gives the SVG accessible content, so Chrome surfaces it inside the labelled button as its own node. From the a11y tree on the demo:

```
button "Skip back ten seconds"
  └─ img
       └─ generic: "10"
```

The announced button *name* is correct either way — `aria-label` wins — so this isn't a naming bug. It just leaves a stray image node that a screen reader can land on in browse mode, saying nothing useful. The other eight icons have no text content and don't show up at all.

`ia-combo-box` already does this for its decorative icons ([`ia-combo-box.ts:429`](https://github.com/internetarchive/elements/blob/main/src/elements/ia-combo-box/ia-combo-box.ts#L429)), so adding `aria-hidden="true"` to the `<svg>` roots would match. Only these two strictly need it; applying it to all ten is probably tidier than remembering which are which.

src/elements/ia-playback-controls/ia-playback-controls.ts
```diff
@@ -0,0 +1,338 @@
...
+  PrevSectionButtonPressed: 'prev-section-button-pressed',
+  NextSectionButtonPressed: 'next-section-button-pressed',
+};
+
+/** How much one press of the speed button moves the playback rate */
+const PLAYBACK_RATE_STEP = 0.25;
+
+/** The rate the speed button wraps back around to once it passes the top */
+const MIN_PLAYBACK_RATE = 0.5;
+const MAX_PLAYBACK_RATE = 2;
```
nsharma123, line 42: Small doc nit: this one comment sits above both constants but only describes `MIN_PLAYBACK_RATE`, leaving `MAX_PLAYBACK_RATE` undocumented — and "the rate it wraps back around to" reads as if it were describing the *ceiling* on first pass, which is the opposite of what it means. Splitting them is clearer:

```ts
/** The slowest rate, and the one the speed button wraps back around to */
const MIN_PLAYBACK_RATE = 0.5;

/** The fastest rate; pressing the speed button at or past it wraps to the minimum */
const MAX_PLAYBACK_RATE = 2;
```

src/elements/ia-playback-controls/ia-playback-controls.ts
```diff
@@ -0,0 +1,338 @@
...
+  /**
+   * The accessible name for the speed button.
+   *
+   * The rate is left bare rather than followed by a unit like "times", because
+   * `@lit/localize` has no plural support, so a unit baked into the string
+   * reads wrong at 1 and translators have no way to correct it. The visible
+   * label beside the button carries the "x".
+   */
+  private get playbackRateLabel(): string {
+    return msg(str`Playback speed, currently ${this.playbackRate}`);
```
nsharma123, line 177: i18n nit, and only worth raising because the surrounding comment reasons so carefully about the pluralisation trap: the rate is interpolated as a raw JS number, so it always formats with a period. In a comma-decimal locale (`de`, `fr`, `es`, …) a translator gets `Playback speed, currently 1.25` and has no way to correct the separator from the message string — same trap as the unit, different axis. The visible `${this.playbackRate}x` label on line 76 has it too.

`Intl.NumberFormat` fixes both:

```ts
private get formattedRate(): string {
  return new Intl.NumberFormat().format(this.playbackRate);
}
```

Fine to leave for the follow-up if the player is English-only today — just flagging it while the reasoning is fresh in the file.

nsharma123 (approved):

src/elements/ia-playback-controls/ia-playback-controls.ts
jbuckner: Fixed in bb5f5c7. Added `padding: 0` to `#play-pause-btn` with a comment on why, so the glyph keeps the same proportion of the circle at every diameter.

src/elements/ia-playback-controls/ia-playback-controls.ts
jbuckner: Fixed in bb5f5c7, using your suggested shape: the rate is clamped on the way in as well as the way out. Added a test that starts at 0.1 and expects 0.75 after one press, which fails without the clamp.

src/elements/ia-playback-controls/ia-playback-controls.ts
jbuckner: Fixed in bb5f5c7. Went with the getter version so the bad value never reaches the screen reader even without a click: everything that displays, announces or steps the volume now reads it through a `clampedVolume` getter. Two tests cover it, one for the negative step and one asserting `volume="-1"` and `volume="loud"` both announce "currently 0 percent".

src/elements/ia-playback-controls/ia-playback-controls.test.ts
jbuckner: Good catch, and a fair one. Split out to WEBDEV-9046 rather than hold the radio player release: https://webarchive.jira.com/browse/WEBDEV-9046

src/elements/ia-playback-controls/assets/replay.ts
jbuckner: Agreed. Tracked in WEBDEV-9046 with the rest of the playback-controls follow-ups: https://webarchive.jira.com/browse/WEBDEV-9046

src/elements/ia-playback-controls/ia-playback-controls.ts
jbuckner: The clamp fix reworded this comment, but you're right that MAX is still undocumented. Tracked in WEBDEV-9046: https://webarchive.jira.com/browse/WEBDEV-9046

src/elements/ia-playback-controls/ia-playback-controls.ts
jbuckner: Tracked in WEBDEV-9046: https://webarchive.jira.com/browse/WEBDEV-9046

PR #82, opened 2026-08-27
WEBDEV-8966: Migrate scrubber-bar into elements

jbuckner:
[WEBDEV-8966](https://webarchive.jira.com/browse/WEBDEV-8966). Stacked on [#80](https://github.com/internetarchive/elements/pull/80), part of [WEBDEV-8962](https://webarchive.jira.com/browse/WEBDEV-8962).

Brings `scrubber-bar` over as `ia-scrubber-bar`, and its child as `ia-section-marker`.

Bugs fixed on the way through:

- Section markers are sorted numerically, on a copy. `.sort()` was called bare, which orders lexicographically, so `[9, 20, 100]` picked the wrong markers either side of the playhead. It also sorted in place, reordering the consumer's own array.
- Keyboard scrubbing counts as interacting. Arrowing along the bar set the value but left the interacting flag false, so the next playback tick snapped the thumb back. Blur releases a latched interaction, and Tab deliberately doesn't count, since its keyup lands wherever focus went.
- `touchcancel` ends an interaction, same as on the waveform.
- `percentage` returns 0 rather than `NaN` when `max === min`.
- The fill, markers and thumb follow the themed track height. They were pinned to numbers that only suit the default 10px track, so theming the height left the fill and thumb floating off it.
- The slider has an accessible name.

Also: `updateWebkitSliderStyle()` is gone. It built a `<style>` block as a string and wrote it in via `innerHTML` on every value change, but the element it styled is an ordinary div, not a track pseudo-element, so a custom property does the same job. Marker modes are computed in `render()` and passed as property bindings rather than assigned by `querySelector`.

One thing deliberately **not** fixed: the marker arrows keep their unconditional `visibility: hidden`. The `mode-*` rules only animate `opacity`, which can't reveal a hidden element, so those arrows have never been visible in production. Making UI nobody has seen suddenly appear didn't feel like a migration's call — happy to do it as a follow-up.

CI shows only `deploy-preview` here because the PR isn't based on `main` ([WEBDEV-8972](https://webarchive.jira.com/browse/WEBDEV-8972)). Verified locally: build, full suite 277 passing, eslint and prettier clean.

[WEBDEV-8966]: https://webarchive.jira.com/browse/WEBDEV-8966?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pezvi (approved):
lgtm

PR #83, opened 2026-08-27
WEBDEV-8967: Migrate expandable-search-bar into elements

jbuckner:
[WEBDEV-8967](https://webarchive.jira.com/browse/WEBDEV-8967). Stacked on [#82](https://github.com/internetarchive/elements/pull/82), part of [WEBDEV-8962](https://webarchive.jira.com/browse/WEBDEV-8962).

Brings `expandable-search-bar` over as `ia-expandable-search-bar`, and its child as `ia-quick-search`.

Fixed on the way through:

- Quick search entries are buttons. They were anchors with a click handler and no `href`, which nothing can focus and no keyboard can activate — the file carried two `eslint-disable lit-a11y/*` comments admitting it. Selecting an entry now also returns focus to the input, since closing the list hides the element that had it and the browser would otherwise drop focus to the body.
- Value tracking binds to `input` rather than `keyup`, so a mouse paste, autofill, a drop and an IME composition all register. Arrow keys no longer report a change that didn't happen. `keyup` still handles Enter.
- The input and both buttons have accessible names, and the disclosure reports `aria-expanded`.
- `placeholder` and `label` are properties running through `msg()`, so they can be translated and overridden instead of being a hardcoded "Search".

`QuickSearchEntry` is an interface, with `data` optional.

The `updated()` hook that writes `searchTerm` into the field looked redundant next to the `.value` binding, but it isn't, and it now carries a comment explaining why: Lit dirty-checks a property binding against the value it last *committed*, not against the DOM, so a consumer that normalises `searchTerm` in its own `inputchange` handler can land back on the committed value and leave the field showing something else. There's a test for it.

CI shows only `deploy-preview` here because the PR isn't based on `main` ([WEBDEV-8972](https://webarchive.jira.com/browse/WEBDEV-8972)). Verified locally: build, full suite 299 passing, eslint and prettier clean.

[WEBDEV-8967]: https://webarchive.jira.com/browse/WEBDEV-8967?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbuckner:
Isa, WEBDEV-8967 has isa-cr-lgtm and isa-qa-lgtm on it, but there's no approval on the PR itself. The radio player stack merges all-or-nothing, so this one is blocking the whole set. Mind clicking Approve?

Heads up that I merged main in just now to clear a conflict in src/elements/index.ts, so the diff moved a little since you looked. Lint is clean and all 615 tests pass on the top of the stack.

bfalling (approved):
Approved by Isa

PR #85, opened 2026-08-27
WEBDEV-8974: Migrate the radio player search handler into elements

jbuckner:
[WEBDEV-8974](https://webarchive.jira.com/browse/WEBDEV-8974). Stacked on [#84](https://github.com/internetarchive/elements/pull/84), part of [WEBDEV-8962](https://webarchive.jira.com/browse/WEBDEV-8962).

Split out of the player's ticket because it's pure logic with no DOM in it, and putting the two together made a ~3,000 line PR. `src/elements/ia-radio-player/` has no element in it yet — that's the next and last PR, which lands beside this.

Seven bugs came over with it, all fixed and pinned by tests that fail without the fix:

- **An empty query hung the browser.** The scan loop is `while ((i = haystack.indexOf(query, i + 1)) !== -1)`, and `indexOf` with an empty needle clamps to the string length and never returns `-1`, so the loop can't end. Confirmed in node — the index sticks at 32 for a 32-character haystack.
- **The transcript index mismeasured music entries.** Spans came from `displayText.length` while `rawText` was what went into the merged string. For a music entry those differ (`[Transcript unavailable]` against a usually-empty `rawText`), so it claimed a 24-character span overlapping the entries after it.
- **The merged string was trimmed after spans were measured against the untrimmed version.** A transcript opening on music starts with an empty entry, so the trim slid every span one character left and search returned nothing for the first entry.
- **A match ending on an entry's last character got the wrong end time.** The span end is exclusive, so the lookup landed on the joining space, found nothing, and fell back to the entry the match started in.
- **Overlapping matches built backwards spans**, and `substring` reverses those silently, so text came back duplicated. Only the first of an overlapping set is taken now.
- **Entries with no text vanished during a search**, taking their time codes with them, so the transcript lost its music rows.
- **The full-text backend built its tags into a `RegExp` unescaped.** The default `{{{`/`}}}` survive only because JS treats an invalid quantifier as a literal; a tag containing a bracket would throw.

`Range` is `TextRange` here — it's re-exported from the package root and shadowing the DOM's `Range` is a trap. `TranscriptEntryRange` and `SearchResult` are interfaces; `TextRange` stays a class for its `length` getter. The response payload types are declared rather than `any`.

CI shows only `deploy-preview` here because the PR isn't based on `main` ([WEBDEV-8972](https://webarchive.jira.com/browse/WEBDEV-8972)). Verified locally: build, full suite 380 passing, eslint and prettier clean, 95% coverage on this subsystem.

[WEBDEV-8974]: https://webarchive.jira.com/browse/WEBDEV-8974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


latonv (approved):
No blockers found, LGTM.

PR #86, opened 2026-08-27
WEBDEV-8969: Migrate radio-player into elements

jbuckner:
[WEBDEV-8969](https://webarchive.jira.com/browse/WEBDEV-8969). Stacked on [#85](https://github.com/internetarchive/elements/pull/85). Last of the eight in [WEBDEV-8962](https://webarchive.jira.com/browse/WEBDEV-8962).

The player itself, plus `ia-search-results-switcher`. Wires up the seven components below it, and uses `ia-status-indicator` in place of `ia-activity-indicator`, which isn't migrating.

Most of the bugs here are one mistake repeated: a falsy guard on a number where zero is a real value.

- **The first transcript line couldn't be clicked** — the one entry that starts at 0. Same guard also meant rewinding to 0:00 left the scrubber where it was, scrubbing to 0% did nothing, and muting was ignored.
- **Backspacing the search box empty left the old term in place**, since an empty string is falsy too.
- **A duration of `Infinity` (a stream) or `NaN` (metadata not loaded) was accepted** as truthy. Only finite durations are now, and the percentage guards a zero duration instead of producing `NaN`.
- **The section skip buttons took `Math.min` of an empty list** at either end of a track. An audio element throws a `TypeError` on a non-finite position (verified in Chrome), which inside a listener surfaces as an uncaught window error, not a visible failure.
- **A failed search left the spinner up for the life of the page.** Nothing put it back. It's in a `finally` now, with a `searchFailed` event so a consumer can say something.
- **Overlapping searches could resolve out of order**, leaving older results on screen under the newer term. Reachable over the network with `FullTextSearchBackend`. Each run is numbered and only the newest reports back.
- **The quick searches were plumbed through but dead.** The disclosure that opens the list was never enabled, and nothing listened for a selection, so the entries could be neither seen nor picked.
- **The switcher's arrows had no accessible name** — that's the open bug [WEBDEV-8926](https://webarchive.jira.com/browse/WEBDEV-8926) — and the counter had nothing saying what it counted. Both fixed. The search outcome is announced through a live region that's always in the DOM, since a region inserted with its text already in place isn't announced.

Every one of those has a test that fails when the fix is reverted.

Also dropped ~60 lines of `-ms-grid-*` CSS. It only ever served IE11, which can't run this package's ES2021 output.

`RadioPlayerConfig` and `MusicZone` are interfaces. The controls' playback mode is a binding off `isPlaying` rather than being assigned into the shadow DOM.

The story generates its own audio (a quiet WAV tone that steps down in pitch), waveform and logo, and wires up a real `SearchHandler` over a `LocalSearchBackend`, so the whole thing is demoable with no network.

CI shows only `deploy-preview` here because the PR isn't based on `main` ([WEBDEV-8972](https://webarchive.jira.com/browse/WEBDEV-8972)). Verified locally: build, full suite 437 passing, eslint and prettier clean, ghpages demo builds.

[WEBDEV-8969]: https://webarchive.jira.com/browse/WEBDEV-8969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WEBDEV-8926]: https://webarchive.jira.com/browse/WEBDEV-8926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bfalling (approved):
LGTM

PR #89, opened 2026-08-27
WEBDEV-8973: Rebuild the demo scroll spy on reconnect

jbuckner:
`AppRoot` tore the scroll spy down on disconnect but could only build it back up from `updated()`, so the sidebar's in-view highlight stayed dead after a reconnect.

The reason it never fired is subtle: `entryFromHash` returns `storyEntries.find(...)`, so an unchanged hash resolves to the *same* object. Lit's default `hasChanged` is `!==`, so re-reading the hash in `connectedCallback` requested no update, `updated()` never ran, and the observer was never rebuilt. Fixed with an explicit `requestUpdate()`, which keeps `updated()` as the single owner of the spy's lifecycle.

app-root is the demo's root element and isn't shipped in the package, so this mainly bites tests and anything that moves the element around the DOM.

Worth knowing about the test: it settles every story load before disconnecting. Without that it passed even with the bug present, because in a cold run the in-flight `_loadStory` calls request their own updates and rebuild the spy by accident. I checked it fails without the fix and passes with it, both on its own and in a full-file run.

https://webarchive.jira.com/browse/WEBDEV-8973

https://claude.ai/code/session_01CUM1e29MLyK3kBDiZYgERx

### DIFF
--- a/demo/app-root.test.ts
+++ b/demo/app-root.test.ts
@@ -27,12 +27,24 @@ function anchorIds(el: AppRoot): string[] {
   );
 }
 
+/** The sidebar link the scroll spy currently marks, by href. */
+function inViewHref(el: AppRoot): string | null {
+  const marked = el.querySelector('.ia-elem-link.in-view');
+  return marked ? marked.getAttribute('href') : null;
+}
+
+function scrollToAnchor(el: AppRoot, id: string) {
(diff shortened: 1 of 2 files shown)


pezvi (commented):
LGTM

PR #91, opened 2026-08-27
WEBDEV-8951 Add a mediatype center icon to ia-status-indicator

iisa:
[WEBDEV-8951](https://webarchive.jira.com/browse/WEBDEV-8951)

**Preview:** https://internetarchive.github.io/elements/pr/pr-91/#elem-ia-status-indicator

Extend and upgrade the status indicator so that it can carry a mediatype icon in the centre of its ring, not just the three animated dots it draws today.

## API

```ts
export type LoadingStatus = 'ready' | 'loading' | 'success' | 'error';

export type MediaTypeIcon =
  | 'audio' | 'collection' | 'etree' | 'images' | 'search'
  | 'software' | 'texts' | 'tv' | 'video' | 'web';
```

| Property | Type | Default |
|---|---|---|
| `mode` | `LoadingStatus` | `loading` |
| `mediatype` | `MediaTypeIcon` | unset |
| `hideDots` | `boolean` | `false` |
| `loadingStyle` | | **removed** |

Slot: `icon` — for glyphs with no mediatype, e.g. item-nav's book.

## Behaviour

**The centre is a mediatype icon or the dots, never both.** In `loading`, first match wins:

1. Slotted `icon`
2. `mediatype` glyph — static, monochrome with the ring
3. Nothing, when `hideDots` is set
4. Three animated dots — *today's default, so existing consumers see no change*

**Two ring geometries, both kept.** 120x120 for dots, a roomier 100x100 for a glyph. Unifying them would shift every existing call site for no gain.

`success` / `error` ignore `mediatype` — they replace the whole SVG, and `error` has no ring.

`ready` renders a placeholder that reserves the indicator's space without drawing, so a consumer's layout doesn't shift once the indicator appears. `ia-otp-form` relies on it for its idle state.

**Unmapped mediatypes degrade to dots.** `account`, `data`, `movies` have no glyph here.

## Breaking changes

Consumer updates are separate tickets: [WEBDEV-8982](https://webarchive.jira.com/browse/WEBDEV-8982) offshoot, [WEBDEV-8983](https://webarchive.jira.com/browse/WEBDEV-8983) bookreader.

* `loadingStyle` removed. Four offshoot sites pass `"ring-dots"` (the default) — drop the attribute. `home-page.ts:215` passes `"ring"` → `hideDots`.

## Implementation

Glyphs are CSS masks, not `` — `` can't be recoloured. They ship as separate `.svg` files under `icons/` to keep path data out of the bundle, grouped the way `ia-item-navigator` groups its own. The mask `url()` stays **quoted**: bundlers inline these as `data:` URIs containing apostrophes, and an unquoted `url()` silently drops the declaration and paints a solid block.

`mask-size: contain` centres a glyph's *viewBox*, not its ink, so `collection`, `texts` and `web` rendered low in the ring — roughly 29% empty above the ink and 6.7% below. Their viewBoxes now give the ink the same 20-unit vertical margins the balanced glyphs already had, which centres it without changing scale. Cropping tight to the ink would also have fixed the alignment, but it scales a glyph up to fill its box.

`MediaTypeIcon` is a local union, not `@internetarchive/field-parsers` — no runtime dep for a type, and it can't promise a glyph we don't ship.

`maskedIcon` moved from `ia-item-navigator` to `src/util` so both components share one copy of the helper and its quoted-`url()` fix, rather than each carrying its own. Only import paths changed for the navigator. Retiring `ia-itemnav-loading-view` in favour of this component is [WEBDEV-8980](https://webarchive.jira.com/browse/WEBDEV-8980).

## Demo

Checking a glyph at a realistic size meant typing values by hand, so the story gained real controls. The additions are generic and available to any story:

* `StyleInputSettings.presets` — one-click values, rendered under the input, or beside it with `presetsInline`. Width carries the widths consumers actually use; Loading carries White.
* `StyleInputSettings.section` / `PropInputSettings.section` — groups consecutive inputs under a heading, giving the Styles panel a `Color` group and the Settings panel an `Accessible titles` group instead of repeating the prefix in every label.
* The Styles panel's own `Revert` is opted into with `revertable`; the story's remaining button resets only the props, which `Revert` doesn't cover.
* Radios apply on change, so the demo and the usage example follow a selection without a trip to Apply.
* Props at their default are left out of the usage example — a consumer only needs to see what they are actually changing.
* The dark surface is derived from the loading colour rather than toggled: a light indicator needs a dark background to be visible at all. This needed `stylesApplied` to bubble and be composed, since it previously stopped at story-template's shadow boundary.

## Testing

**361 tests pass repo-wide**, up from 331. `madge --circular` clean, build clean, lint clean. The 10 glyphs land in `dist/src/elements/ia-status-indicator/icons/`.

`ia-status-indicator.ts` is at 100%. Total coverage moved 88.5% to 92.1%.

The riskiest parts of this work are visual, so they are asserted in the real browser rather than eyeballed:

* `icons.test.ts` — every shipped glyph's ink must sit symmetrically inside its viewBox. This is the regression lock for the centring fix above.
* `ia-status-indicator.test.ts` — a `rendered appearance` block: the glyph is centred, stays inside the ring, paints the same colour as the ring, and scales with the indicator's width.
* `ia-status-indicator-demo.test.ts` — drives the story's own controls and asserts what the indicator renders. Nothing reaches into the settings panels' internals, but exercising them this way took `story-prop-settings` from 41.9% to 88.4% and `story-template` from 87.5% to 96.2%.

Each was checked by breaking what it guards — reverting a glyph's viewBox fails the margin test, oversizing the glyph fails containment, removing the radios' change handler or either empty-string guard fails the demo tests.

## QA

Open the preview at `#elem-ia-status-indicator`.

**Component**

- [ ] Regression first: all defaults → 120x120 ring with three animated dots, unchanged from the deployed demo
- [ ] Each of the 10 mediatypes → glyph centred, sized, not clipped; dots gone; ring becomes 100x100
- [ ] Glyph is static while the ring spins
- [ ] `Hide dots` true with mediatype `none` → bare ring
- [ ] `Mode` success, then error, with a mediatype set → mode glyph wins, no centre glyph
- [ ] Shadow root: glyph is `aria-hidden`, exactly one `<title>`

**Styles panel**

- [ ] `Width` presets sit under the input; one click applies, no Apply
- [ ] Presets match real consumers: Default 1.25rem, OTP form 3rem, Page 4rem, Theater 5rem, Account settings 6rem
- [ ] Glyph scales with width, stays centred
- [ ] `Color` heading groups Loading / Success / Error
- [ ] Loading's `White` preset sits beside the swatch, not under it
- [ ] `White` → glyph and ring both white, and the demo gains a black background so they stay visible
- [ ] Set Loading back to a dark colour → black background goes away
- [ ] **If the glyph paints as a solid block or won't recolour, the mask `url()` quoting has regressed** — check computed `mask-image`, not just the visual
- [ ] `Revert` → all four inputs back to defaults, background off

**Settings panel**

- [ ] Any radio updates the demo and the usage example immediately, without Apply
- [ ] `Accessible titles` heading groups Loading / Success / Error at the bottom; text fields, still need Apply
- [ ] Usage example omits props at their default — all defaults collapses to `<ia-status-indicator></ia-status-indicator>`
- [ ] Usage example shows property bindings:

```html
<ia-status-indicator
  .mode=${'error'}
  .mediatype=${'texts'}
  .hideDots=${true}
></ia-status-indicator>
```

- [ ] `Reset props` → mode, mediatype, hideDots back to defaults; example back to the bare tag

**Elsewhere**

- [ ] `#elem-ia-otp-form` — flip Validation Status idle → loading, the input row must not jump
- [ ] Safari — ring still spins (the explicit `100%` keyframe exists because Safari needs it)

[WEBDEV-8951]: https://webarchive.jira.com/browse/WEBDEV-8951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[WEBDEV-8980]: https://webarchive.jira.com/browse/WEBDEV-8980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[WEBDEV-8982]: https://webarchive.jira.com/browse/WEBDEV-8982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/elements/ia-status-indicator/ia-status-indicator.ts
```diff
@@ -21,29 +85,77 @@ export class IAStatusIndicator extends LitElement {
...
-
   /* The state of the indicator that should be shown */
-  @property({ type: String }) mode: LoadingStatus = 'loading';
+  @property({ type: String, reflect: true }) mode: LoadingStatus = 'loading';
+
+  /* An optional mediatype glyph to render in the middle of the loading ring */
+  @property({ type: String, reflect: true }) mediatype?: MediaTypeIcon;
+
+  /* Renders a bare ring, with no dots in the middle. Ignored when a center icon is present. */
+  @property({ type: Boolean, reflect: true }) hideDots = false;
```
jbuckner: How come you're reflecting these 3 properties?

src/elements/ia-status-indicator/ia-status-indicator.ts
```diff
@@ -1,14 +1,78 @@
...
 /**
- * Renders an SVG indicator, which defualts to an animated circular indicator
+ * The states this indicator renders a real indicator for.
+ *
+ * Any other value renders the invisible placeholder instead, which reserves the
+ * indicator's width without drawing anything. That fallback is relied upon by
+ * consumers with an idle/pre-validation state (see `ia-otp-form`), which keeps
+ * their layout from shifting when the indicator appears.
+ */
+export type LoadingStatus = 'loading' | 'success' | 'error';
```
jbuckner: Why did we remove the `ready` status? Since this is a typed value on the property, it should be validating the type that it gets set to

src/elements/ia-status-indicator/ia-status-indicator.ts
```diff
@@ -21,29 +85,77 @@ export class IAStatusIndicator extends LitElement {
...
+    return html`${choose(
+      this.mode,
+      [
+        ['loading', () => this.loadingIndicatorTemplate],
+        ['success', () => this.successIndicatorTemplate],
+        ['error', () => this.errorIndicatorTemplate],
+      ],
+      // Anything unrecognised (e.g. an idle 'ready' state) reserves space
+      // without drawing. See the note on LoadingStatus.
+      () => this.placeholderTemplate,
```
jbuckner: When would this get reached? `this.mode` is of type `LoadingStatus` so it should only ever be one of those types.

src/elements/ia-status-indicator/icons/web.svg
jbuckner, line 1: Would cropping the whitespace around this icon fix the vertical alignment issue?

src/elements/ia-status-indicator/ia-status-indicator.ts
```diff
@@ -219,3 +412,9 @@ export class IAStatusIndicator extends LitElement {
     ];
   }
 }
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'ia-status-indicator': IAStatusIndicator;
+  }
+}
```
jbuckner, line 397: We shouldn't need this

src/elements/ia-status-indicator/icons/audio.svg
jbuckner, line 1: Can we organize the icons in a subdirectory so we don't have a bunch of icons at the top-level of the component?

src/elements/ia-status-indicator/ia-status-indicator.ts
iisa: good call, restored in 25e9c7d. i'd taken it out thinking nothing used it, which was wrong — `ia-otp-form` renders the indicator unconditionally and relies on `ready` to reserve the row's width so it doesn't jump when validation starts. back in the union as a real case, and otp-form is back to plain `LoadingStatus`.

_— Claude Code, o/b/o Isa HV_

src/elements/ia-status-indicator/ia-status-indicator.ts
iisa: it wouldn't. that branch only existed to catch `ready` after i'd removed it from the type, so the type and the runtime disagreed. restoring `ready` made it an explicit case in `choose` and the fallback is gone — along with the type widening and comments that were propping it up. net -24 lines.

_— Claude Code, o/b/o Isa HV_

src/elements/ia-status-indicator/ia-status-indicator.ts
iisa: not needed here, so taken away in `bbb825c`. nothing in the repo styles or queries on those attributes yet, so it was speculative on my part — easy to add back if a consumer wants `[mode='error']` as a hook later.

setting from markup still works either way, since `type: String` observes the attribute; what's gone is just the outbound echo when something changes the property in js.

_— Claude Code, o/b/o Isa HV_

src/elements/ia-status-indicator/icons/audio.svg
iisa: done in `3be2711`. matches how item-nav groups its fourteen glyphs — the flat layout only really works for the one- or two-icon components like `ia-combo-box`.

_— Claude Code, o/b/o Isa HV_

src/elements/ia-status-indicator/icons/web.svg
iisa: good catch — it was three glyphs, not just `web`. `collection` and `texts` had the same thing: wide-short ink sitting against the bottom of a 300x300 box, ~29% empty above and 6.7% below, and `mask-size: contain` centres the box rather than the ink.

fixed in `35cad3c` by giving those viewBoxes the same 20-unit vertical margins the balanced glyphs already have. cropping tight to the ink would also have fixed the alignment, but it scales every glyph up to fill its box (86.7% -> 100%) and starts crowding the ring, so padding to symmetry keeps the scale identical. margins are now 8.5-8.8% top and bottom, ink still 86.7% of the box — same as `software`, which was already right.

_— Claude Code, o/b/o Isa HV_

src/elements/ia-status-indicator/ia-status-indicator.ts
iisa: i left this in deliberately — the README asks for it: ["Declare each element in `HTMLElementTagNameMap` so `querySelector` is typed and a mistyped tag in a template is caught at build time"](https://github.com/internetarchive/elements/blob/6d4f44006c034ddd7dffe281ac0843345f9d3763/README.md#L157). item-nav follows it in seven files (`ia-item-navigator`, the four `ia-itemnav-*` elements, and the two menu panels).

dropping it here would make the status indicator the one component that doesn't. happy to take it out if you'd rather retire the rule, but then the README and item-nav should probably go with it — did you mean something narrower?

_— Claude Code, o/b/o Isa HV_

jbuckner (approved):

PR #92, opened 2026-09-02
WEBDEV-9009: Fix the demo scroll spy test on a taller demo page

jbuckner:
The otp-input story takes focus as it loads, which scrolls the page. The test asserted on the scroll spy before waiting for the stories to settle, so the first anchor was already out of view.

Green on main either way. It only fails once the demo has enough elements to scroll, which is the radio player stack.

https://claude.ai/code/session_017RuJpnAKn9TtFvMxcEgwdm


pezvi (approved):
lgtm

PR #93, opened 2026-09-04
WEBDEV-9022: Migrate the image viewer into elements

jbuckner:
Ports offshoot's image viewer in as `ia-image-viewer`, split into a slide element, a controls element, and a reactive controller for touch/wheel/keyboard. Sub-elements are namespaced `ia-imgview-*` so they can't collide with offshoot's copy, which still registers `image-viewer` verbatim.

No dependencies added. It took metadata-service `File` objects and an `ItemDetails` only to build download URLs, so it now takes a plain `{ name, url, title? }` list. It also called `history.replaceState` and reached for offshoot's analytics handler; those are events now (`imageChanged`, `imageLoaded`, `imageLoadFailed`), so the host decides.

Reduced motion moves from CSS to JS, which fixes two bugs it was hiding: `animation: none` on the wrap glow meant `animationend` never fired, so the glow sat at full opacity forever, and the swipe settle set its transition inline from JS where the media query couldn't reach it. Navigation now lands immediately with nothing to wait on.

Also folds in WEBDEV-9023: the image link is a real anchor rather than a `role="link"` div, the nav labels go through `msg()`, and a live region announces the image as it changes.

Offshoot swaps over in WEBDEV-9024, against a prerelease of this.

https://claude.ai/code/session_01BmPnn4m5EKaMYPjNvWqpJp


pezvi (approved):
lgtm

PR #94, opened 2026-09-04
WEBDEV-9025: Add a CLAUDE.md covering the publish flow

jbuckner:
The README's publishing section is written for the web UI ("Tap `Draft a new release`") and doesn't say what the pre-release checkbox does, so this writes down the bits that only live in `npm-publish.yml`: the prerelease flag is what picks the `alpha` vs `latest` dist-tag, `package.json` holds the version that actually gets published (not the tag name), and publishing is OIDC trusted publishing with no token so it can't be done from a laptop. Plus the `gh release create` commands and the worktree/merge conventions.

Everything in it was checked against the repo, the tags, the workflow run history and the npm registry.

Worth knowing separately: `0.2.12` is npm's current `latest` but has no `v0.2.12` tag, no GitHub release, and no provenance attestation, so it was published by hand. Recorded on the ticket, not fixed here.

WEBDEV-9025: https://webarchive.jira.com/browse/WEBDEV-9025

https://claude.ai/code/session_01DaEr5YT1xpjK5ydNZPvjwx

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -48,8 +48,10 @@ The workflow bumps nothing. It checks out the released commit and publishes
 whatever `package.json` says, so the tag name and the version have to agree.
 
 Always bump with `pnpm version`, which edits `package.json`, commits, and creates
-the matching tag in one step. The tag gets a `v` prefix from npm's default
-`tag-version-prefix`. Never hand-write a tag.
+the matching tag in one step. The commit message is the bare version (`0.2.13`)
+while the tag carries a `v` prefix (`v0.2.13`) from npm's default
+`tag-version-prefix`, so the prefix comes from npm, not from the commit. Never
+hand-write a tag.
 
 ### Prerelease, from a feature branch
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,120 @@
+# elements
+
+A web component library from the Internet Archive, published as
+`@internetarchive/elements`. See [README.md](README.md) for installation, usage,
+adding a component, styling, and the component inventory.
+
+Node 24+ and pnpm 11+. Use pnpm, never npm or yarn.
+
+```zsh
+pnpm install
+pnpm run dev     # demo at the vite dev server
+pnpm run test    # vitest in real chromium
+pnpm run lint    # eslint + prettier --check, and CI enforces it
+```
+
+## Publishing
+
+Publishing is driven entirely by **GitHub Releases**. Publishing a release fires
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,122 @@
...
+merged, reviewed code or not at all.
+
+The version bump is its own PR, titled after the version (`v0.2.12 (#74)`). Once
+that has merged:
+
+```zsh
+git checkout main && git pull
+pnpm version [major | minor | patch]
+git push && git push --tags
+gh release create v0.2.13 --latest --generate-notes --title v0.2.13
```
latonv: Running `pnpm version` will by default edit `package.json` and create a new commit. If this sequence of commands is run after merging in a version bump via its own PR, it would be pushing up a _further_ version bump, directly to the `main` branch, which I doubt is what we want.

Q: Does our gh release flow automatically create a tag? And if so could we avoid the step of running `pnpm version` on `main` manually, instead relying on the earlier version bump PR and auto-tag?

CLAUDE.md
```diff
@@ -0,0 +1,122 @@
...
+  itself both resolve the tag server-side.
+- **A version can never be republished.** npm rejects a duplicate, so a botched
+  publish needs a new version number rather than a retry.
+- **`canary` and `rc` dist-tags are stale.** They point at old prereleases from
+  manual publishes. The workflow only ever writes `alpha` and `latest`. Don't add
+  more dist-tags.
+- **Every final release must be tagged.** The tag is the only durable pointer from
+  an npm version back to its source commit. `0.2.12` is the cautionary case: it is
+  the current `latest` with no tag, no GitHub release, and no provenance, because
+  it was published by hand.
```
latonv, line 118: Ah, this one was my bad. I must have forgotten that we had a more automated process for releases here! 🤦

latonv (commented):
Just noted a potential contradiction b/w the rule about using PRs for version bumps and the example commands that seem to skirt this. Otherwise the rest looks good!

CLAUDE.md
jbuckner: **Yep, good catch. Dropped `pnpm version` from the main step.** The bump PR already carries the version, so once it lands main just gets `git tag` + `gh release create`.

To your Q: yes, publishing the release is what fires `npm-publish.yml`, and `gh release create` will make the tag itself if it's missing. Kept the explicit `git tag` step anyway so the release pins to the commit you've actually got checked out rather than whatever main is server-side at that moment. Reworded the gotcha to say that.

latonv: Updated version looks good. Thanks for the fix!

latonv (approved):

PR #106, opened 2026-09-15
WEBDEV-9066: Guard custom element registration against duplicate defines

jbuckner:
Importing two element subpaths on the same page throws and one of them never registers. Repros on the published 0.4.0 in headless Chromium:

```
Failed to execute 'define' on 'CustomElementRegistry': the name "ia-status-indicator" has already been used with this registry
```

Each subpath esm.archive.org serves is its own self-contained bundle, so a shared component is built into every entry point that imports it. Two subpaths means two copies and two `customElements.define()` calls for the same tag. The second throws mid-evaluation, which aborts the rest of that bundle, so it's partway through: some tags from the failed bundle register and the rest are lost. Bundler consumers like offshoot never see it because Vite dedupes to one instance.

12 tags are defined by more than one entry point. `ia-status-indicator` is in `ia-button`, `ia-dropdown-search-bar`, `ia-otp-form`, `ia-radio-player` and its own subpath, so any two of those collide.

So this adds `src/util/custom-element.ts`, a `@customElement` that skips the define when the tag is already claimed, and points every call site at it. First copy wins. That's safe here: nothing does `instanceof` on a shared element class and the only class-level imports of one are `import type`, so the copies are interchangeable. There's an eslint rule to keep Lit's version from coming back.

The demo and story files are swapped too. They aren't published so they can't hit this, but leaving them on the Lit decorator just invites someone to copy the wrong import into a new element.

### What I didn't do

There's a second, structural fix: a single-URL entry point would give non-bundler consumers one module instance and no duplicates at all. That's a bigger change and it argues with #81, which drops the package root. Skipping it for now since petabox usage is getting deprecated anyway. The guard is worth having on its own for any non-bundler consumer.

### Verifying

Unit tests cover the guard, including that registration keeps working after a duplicate is skipped, which is the part that actually breaks.

The real failure mode needs two separately-built bundles, which no unit test can produce, so I checked that by hand: esbuild `dist/.../ia-dropdown-search-bar` and `dist/.../ia-radio-player` into two self-contained bundles and load both on one page.

- Without the guard: radio-player throws, doesn't register, doesn't upgrade, and `ia-transcript-view` and `ia-waveform-progress` are lost with it.
- With it: both import, all six tags register, both elements attach a shadow root.

694 tests pass, lint 0 errors, prettier clean.

https://webarchive.jira.com/browse/WEBDEV-9066

https://claude.ai/code/session_01PoknqYwGpzkxFE3kfcYW81


src/util/custom-element.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @param tagName The tag name to register the element under.
+ */
+export function customElement(tagName: string) {
+  return (
+    classOrTarget: CustomElementClass,
+    context?: ClassDecoratorContext<new () => HTMLElement>,
+  ): void => {
+    const define = (): void => {
+      if (customElements.get(tagName)) return;
```
latonv: Q: Would it perhaps be sensible to emit some kind of debug info when we ignore a custom element whose name already exists? I'm thinking of cases where, e.g., an older version of a package gets defined before a newer one comes in, where it would be handy to know upfront rather than have to dig into why we're not seeing the newer behavior.

latonv (approved):
LGTM, just had one thought about possible logging when we skip a define.

src/util/custom-element.ts
jbuckner: Good call, added in e1df5b2. It warns with the tag name and the two things it could mean.

Worth saying that the skew case you're describing is worse than the one this PR set out to fix. Before this, two versions on a page threw, so you'd at least notice. The guard makes it quiet, and quiet-and-wrong beats loud-and-broken for availability but it's much harder to debug. So the warning isn't really a nicety here, it's covering a hole the guard opens.

I can't tell the cases apart in code, which is the annoying part. Two builds of the same element differ in nothing observable at that point, and a version bump that only changes rendering wouldn't show up in the class either. So the message names the tag, says the first registration wins, and points at the two things to go look for rather than guessing which one it is.

It also fires when an unrelated element already owns the tag, e.g. a host page with its own `ia-button`. My first cut said "another copy of it was skipped" and called it "Expected", which would have been actively wrong in that case, so the wording covers all three now.

One thing I tried and dropped: I had it deduped per tag with a module-level Set. That's useless in production. Each bundle copy has its own module instance and its own empty Set, and within one copy a tag only ever gets defined once, so it never dedupes anything. It only appeared to work in the tests, where everything shares one instance. Dropped it, so it's now one warning per skipping copy, which is the right granularity anyway.

Verified against two real bundles from `dist` on one page: exactly one warning for `ia-status-indicator`, everything still registers.

PR #107, opened 2026-09-16
WEBDEV-9063: Typecheck in CI

jbuckner:
Nothing in CI checks types. `vitest` transpiles without checking and eslint isn't type-aware in this config, so a type error passes CI and then turns up at `pnpm build`, which is publish time.

WEBDEV-9050 is what prompted it: a private getter named `ariaValueNow` collided with `ARIAMixin` and cascaded into 22 errors across two files, with lint green and all 650 tests passing.

So this adds a `typecheck` job running `tsc --noEmit`, plus a `pnpm run typecheck` for local use. Its own job rather than a step on `lint`, so when it fails the checks list says typecheck instead of making you open the lint job to find out. It covers `demo/` as well as `src/`, since that's what tsconfig includes.

### Verifying

Re-injected the 9050 bug (a private `ariaValueNow` getter on `ia-status-indicator`) and ran all three checks against it:

- `pnpm lint` — 0 errors
- `pnpm test` — 697 passing
- `pnpm typecheck` — 39 errors, first one `TS2344` on `HTMLElementTagNameMap[K]`

So the two existing checks pass it through and the new one catches it. Clean on `main` as-is.

https://webarchive.jira.com/browse/WEBDEV-9063


pezvi (approved):
lgtm

PR #112, opened 2026-09-21
WEBDEV-9121: Migrate histogram-date-range into elements

jbuckner:
Ports @internetarchive/histogram-date-range in as ia-histogram-date-range, same pattern as the radio-player migration (WEBDEV-8962). Behavior is unchanged except for the actual point of this ticket: the loading spinner switches from the standalone @internetarchive/ia-activity-indicator package to elements' own ia-status-indicator. That package is what caused the duplicate customElements.define collision worked around with a version pin in WEBDEV-9119. This removes it from the dependency graph entirely instead of pinning around it. Follow-up: WEBDEV-9122 drops that pin once collection-browser adopts this.

Added dayjs as a new elements dependency. fix-first-century-years.ts needed a hand-written type for a few dayjs internals the newer version's public types don't declare, though they still exist at runtime.

Test suite ported from mocha/chai to vitest, matching elements' own convention. Dropped the a11y-audit test since chai-a11y-axe isn't used anywhere else here.

Also found and fixed a real bug along the way: willUpdate() was resetting an internal "did the date range actually change" tracker on every date pick, so the debounced histogramDateRangeUpdated event never fired. Reproduces 100% under elements' Lit 3.3.3; the standalone package pinned Lit 2.8 and apparently never hit it. Details in the commit message.

Verified: full elements suite green (34 files, 724 tests), typecheck clean, lint clean, build succeeds.

https://claude.ai/code/session_01Vc5iAyrMu7daxQkYQgWUEf


pezvi (approved):
lgtm

PR #114, opened 2026-09-21
WEBDEV-9130: Exclude nested dist and worktree tests from local runs

jbuckner:
`test.exclude` was `['node_modules', 'dist']`, and neither pattern has a glob, so they only matched the top-level directories. Every worktree under `.claude/worktrees/` got its tests collected on top of your own, both their sources and their built `dist/**/*.test.js`, which fail to import.

With 15 worktrees checked out, `pnpm exec vitest list --filesOnly` went from 2440 collected files to 33.

CI has no worktrees so it never saw this.

https://claude.ai/code/session_01LeNjfWd3FUbZ7kEJMFW1w1


pezvi (approved):
lgtm

